PR #2, opened 2017-01-30
Azure Serverless plugin

pragnagopa:


pragnagopa:
@worldsoup @ac360 - This is the initial PR which has the core pieces. Once we get this in, will send updates with improvements.

pragnagopa:
Adding @yochay as FYI

charpeni:
@pragnagopa what happened with the semicolons?

pragnagopa:
Fixed semicolons. Auto fixing via Eslint removed them

package.json
```diff
@@ -0,0 +1,96 @@
...
+    "azure functions",
+    "iot",
+    "internet of things",
+    "serverless.com"
+  ],
+  "license": "MIT",
+  "main": "index.js",
+  "name": "serverless-azure-functions",
+  "optionalDependencies": {},
+  "readme": "# serverless-azure-functions\r\nWIP – Coming soon...\r\n",
```
christopheranderson: Do we need this if we have the readmeFilename?

shared/parseBindings.js
```diff
@@ -0,0 +1,39 @@
...
+  settings: 'settings',
+  name: 'name',
+  displayName: 'displayName',
+  type: 'type'
+};
+
+module.exports = {
+  getBindingsMetaData: function (serverless) {
+    serverless.cli.log(`Parsing Azure Functions Bindings.json...`);
+    var bindingDisplayNames = [];
```
christopheranderson: Most of these `var`s could be `const`

shared/utils.js
```diff
@@ -0,0 +1,176 @@
...
+
+      if (bindingTypeIndex < 0) {
+        throw new Error('Binding not supported');
+      }
+
+      bindingSettings = parsedBindings['bindingSettings'][bindingTypeIndex];
+      bindingSettingsNames = parsedBindings['bindingSettingsNames'][bindingTypeIndex];
+
+      if (azureSettings) {
+        for (var j = 0; j < Object.keys(azureSettings).length; j++) {
```
christopheranderson: Loops should use `let` instead of var unless you mean to use the variable outside the block scope

package.json
```diff
@@ -0,0 +1,96 @@
...
+  "author": {
+    "name": "pragnagopa"
+  },
+  "dependencies": {
+    "async": "^2.1.2",
+    "azure-arm-resource": "1.6.1-preview",
+    "bluebird": "^3.4.6",
+    "chalk": "^1.1.3",
+    "lodash": "^4.16.6",
+    "ms-rest-azure": "*",
```
christopheranderson: not safe to use `*` versions

christopheranderson:
You ought to be able to set your ESLint settings to warn you about using `var` - should prefer const or let if you need to assign to a different object.

Also, since we're targeting es6, we might want to add an "engine" version to prevent folks from using older Node versions - ideally folks are using 6.5.0 given that's what we're using in Azure right now.

Brygghuset:
If you need help fixing the code review remarks above or anything else, please let me know. I would love to help with anything to get proper Azure support in Serverless.

package.json
mgarrettm: I think the request module is missing from this list, too.

provider/azureProvider.js
```diff
@@ -0,0 +1,577 @@
...
+class AzureProvider {
+  static getProviderName () {
+    return constants.providerName;
+  }
+
+  constructor (serverless) {
+    this.serverless = serverless;
+    this.provider = this;
+    this.serverless.setProvider(constants.providerName, this);
+    subscriptionId = process.env[this.serverless.service.provider.subscriptionId];
```
mgarrettm: This doesn't seem correct. It seems like this used to pull credentials from serverless.yml and then switched to taking them from environment variables, but now instead it kind of does half of both. Environment variables seem better than serverless.yml, because at least it's out of the source, but it's still a worse pattern than the other plugins have. I'm not sure what options there with Azure credentials, though. Anyways, the way the README currently reads, it seems like

```javascript
subscriptionId = process.env['azureSubId'];
servicePrincipalTenantId = process.env['azureServicePrincipalTenantId'];
servicePrincipalClientId = process.env['azureservicePrincipalClientId']; // The 's' should probably get capitalized in the README
servicePrincipalPassword = process.env['azureServicePrincipalPassword'];
```

would be more correct.

pragnagopa:
@mgarrettm - For now using the environment variables for azure credentials, but we are looking into other options to login - command line login  , providing a script to setup credentials.

mgarrettm:
@pragnagopa - Both of those directions sounds promising, thanks for the update. And thanks for the plugin, Serverless support for Azure is really awesome to see!

PR #19, opened 2017-02-22
Fix instructions, add license and contributing

### LINKED ISSUE #18: Update readme
Readme isn't in a great state
 - [x] environment variables aren't correct. 
 - [x] credentials instructions could be more full
 - [x] should point at the boiler plate and use the "install -url" command

christopheranderson:
fixes #18 

- Fixed instuctions
   - Changed install command
   - Expanded/corrected credentials section
   - Added (optional) part to remove

 - Added Contributing section

 - Added LICENSE section

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,41 +4,91 @@ This plugin enables Azure Functions support within the Serverless Framework.
 
 ## Getting started
 
+### 1. Get a Serverless Service with Azure as the Provider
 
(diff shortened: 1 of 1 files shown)


christopheranderson:
Note: we need to publish to npm again to update readme there with new instructions. Need to rev the package version to 0.1.1 before we do that, so I'll wait until we're about to ship to make sure there aren't any other patches we need to take.

Issue #46, opened 2017-03-19
Add CLI option to not clear the existing functions from the hosting environment when deploying


aidanmorgan:
If I want to break my application into a YAML file per aggregate root (partitioning the file by data type) then I have to deploy to different hosts.

Again, this is a perfectly sensible default, however in the effort to save money I would like to be able to deploy them onto the same hosting environment.

I suggest that we add a `--preserve` CLI option for the deploy targets that will skip the deleting of functions that are not declared in the serverless.yml file.

Again - happy to create a PR and submit the change if interested.

Issue #44, opened 2017-03-18
Allow specification of resource group in serverless.yaml


aidanmorgan:
The current experience of automatically creating an Azure resource group for each function is a great "out of the box" experience, however when you are deploying a lot of different functions it would be nice to be able to specify the name of the resource group to use in the serverless.yaml file.

Perhaps something like this:

service: my-example-service
x-azure-resource-group: special-resource-group

I am happy to submit a PR with this change if people are okay with the suggestion.

aidanmorgan:
Alternatively this might be better placed in the provider configuration in the serverless.yaml file?

provider:
  name: azure
  location: Australia East
  resource-group: special-resource-group

lostintangent:
@aidanmorgan This suggestion sounds great to me!

aidanmorgan:
I have implemented it as a resourceGroup item under the provider definition and it seems to work quite well.

cmatskas:
@aidanmorgan is this in one of the pending PRs? I'd love to use this as I think hardcoding the RG name is a bad idea or at least having an option to override would be much better.

aidanmorgan:
@cmatskas I have an implementation of it on a branch, however I am still not comfortable with the default behaviour being to delete the resource group - i think RG's should probably be deleted manually.

There's also some cross-cutting concerns with other functions that are deployed, the implementation I have is not fully tested yet - hopefully I should have some time this weekend to finish it off.

cmatskas:
@aidanmorgan you mean when you run the `serverless remove` command? Yes, this is a grave issue indeed. The needs to be a more granular to allow to either blast the whole thing or remove just the function within the RG. You've already raised this in #48 which is great so hopefully we'll see these changes soon

aidanmorgan:
@cmatskas there is also an issue with "replacing" an existing function, the codebase has assumptions that there is only one function deployed in the function app, so it deletes everything - so I've had to rework a lot of that logic as well.

cmatskas:
@aidanmorgan is this tracked separately? I believe this is an important addition. Devs will need control at resource group, function app (resource) and function level as well

aidanmorgan:
I originally submitted the code in 1 PR, but have cancelled that one and will be submitting several smaller ones - will hopefully get some time on Saturday to get it finished off.

cmatskas:
@aidanmorgan good luck :)

aidanmorgan:
[...]

PR #47, opened 2017-03-20
Changes for Issues 44, 46 and 48

aidanmorgan:
This pull request contains changes for:

#46 
- set the CLI options on the plugin manager to pass the options through to the plugins
- set a config property (`preserveDeployedFunctions`) if a `--preserve` command line argument is provided
- skip deleting the existing functions that are deployed to a host if the `preserveDeployedFunctions` is set

#44 
- allow the specification of a `resourceGroup` property under the provider node in `serverless.yml` file
- updated `AzureProvider` to check for this property and override the default name if it is set.

### DIFF
--- a/deploy/azureDeploy.js
+++ b/deploy/azureDeploy.js
@@ -20,10 +20,6 @@ class AzureDeploy {
       createFunctions
     );
 
-    if(this.options.preserve) {
-      this.serverless.config.preserveDeployedFunctions = true;
-    }
-
     this.hooks = {
       'before:deploy:deploy': () => BbPromise.bind(this)
         .then(this.loginToAzure)
--- a/deploy/azureDeploy.js
+++ b/deploy/azureDeploy.js
@@ -20,6 +20,10 @@ class AzureDeploy {
       createFunctions
     );
 
+    if(this.options.preserve) {
+      this.serverless.config.preserveDeployedFunctions = true;
+    }
+
     this.hooks = {
       'before:deploy:deploy': () => BbPromise.bind(this)
         .then(this.loginToAzure)
--- a/deploy/azureDeployFunction.js
+++ b/deploy/azureDeployFunction.js
@@ -10,6 +10,10 @@ class AzureDeployFunction {
     this.options = options;
     this.provider = this.serverless.getProvider('azure');
 
+    if(this.options.preserve) {
+      this.serverless.config.preserveDeployedFunctions = true;
+    }
+
     Object.assign(
       this,
       loginToAzure,
--- a/deploy/lib/CreateResourceGroupAndFunctionApp.js
+++ b/deploy/lib/CreateResourceGroupAndFunctionApp.js
@@ -1,9 +1,10 @@
 'use strict';
 
 module.exports = {
-  CreateResourceGroupAndFunctionApp () {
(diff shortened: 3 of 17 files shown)


lostintangent:
This is awesome! Quick thought: will the "remove" command behave as expected after your resourceGroup change? Presumably, it would delete your entire RG, which wouldn't be ideal if that RG contained resources that are external to the current Serverless service.

aidanmorgan:
That's a good point, and a side-effect that I hadn't considered. I would suggest that we don't want to delete the resource group in this case.

Would a better change be to change the remove logic to use the `cleanUpFunctionsBeforeDeploy` rather than deleting the actual resources?

I'd say there's probably a difference of intent between wanting to remove a service vs. deleting everything. 

I've not had much experience with the AWS implementation, but the [documentation](https://serverless.com/framework/docs/providers/aws/cli-reference/remove/) reads as if it just removes the service, not the whole deployment.

My suggestion would be to change remove to do the 'cleanUpFunctionsBeforeDeploy' style logic, and add another CLI option to actually drop everything?

lostintangent:
That proposal sounds great. I wonder whether we could determine whether to delete the entire RG based on if the service specified a RG name in the YAML setting, and just clear the functions otherwise. That way, the "getting started" experience is still simple and cleans up entirely after itself, while still allowing more flexibility for more mature projects.

That solution would prevent the ability to customize the RG name that is auto-created, while still having the "remove" command delete the entire RG. But that scenario seems fairly low value, and we could always address it later if we heard feedback.

aidanmorgan:
@lostintangent I've implemented the behaviour as discussed. I also created another issue #48 to describe the change (just for tracability).

aidanmorgan:
Only last change I would like is the ability to use the Serverless variables in the name of the resource group, that way it would be possible to deploy things with names that match the stage.

e.g: 

```
provider:
  name: azure
  location: Australia East
  resourceGroup: example-serverless-rg-${opt:stage}
```

I think this is important as it allows developers to separate their development, integration and production environments. 

I cant seem to find the right place to get the variable substitution to work @lostintangent  do you have anu ideas?

aidanmorgan:
Okay I have resolved the issue with the variable substitution not working, it appears to be caused by the provider's constructor getting called before the Serverless framework has actually completed it's initialisation process.

I moved the logic in the constructor that was deriving values into an initialise() method and now all variables appear to be substituting correctly.

aidanmorgan:
I have also added the WEBSITE_CONTENTAZUREFILECONNECTIONSTRING and WEBSITE_CONTENTSHARE app settings to the ARM templates as the Azure Functions portal was complaining that these values were not set.

pragnagopa:
@aidanmorgan - Thanks for the feedback and the PR. Will discuss the design changes and proposals with @christopheranderson and get back to you. 
Also, please  submit PR against dev branch.

aidanmorgan:
@pragnagopa would you prefer that I separated out each of the issues into their own PR's?

aidanmorgan:
Superceded by #54

pragnagopa:
@aidanmorgan  - I am closing this PR. Please go ahead and submit individual PRs against dev branch when you get a chance.

Issue #53, opened 2017-03-25
Provider is created too early in Serverless process, not all variables have been substituted yet


aidanmorgan:
I would like to be able to have some of the setting values derived from the stage option, however at the moment the AzureProvider instance is created too early on in the initialisation of the framework, which means I can't access these options - additionally the framework has not performed variable substitutions yet (as per [Serverless Variables](https://serverless.com/framework/docs/providers/aws/guide/variables/)).

fiveisprime:
@pragnagopa looks like this was this fixed in #54 and #56. Can this be closed or
were you still investigating potential changes to initialization?

aidanmorgan:
@fiveisprime if I understand correctly this has been merged to dev already

PR #54, opened 2017-03-25
Fix for Issue #53

aidanmorgan:
This is a fix for Issue #53 

The constructor for the AzureProvider performs initialisation of variables used later in the process, however this constructor is called before the Serverless framework has initialised all of the variables from the `serverless.yml` file.

This change moves the initialisation process to each command to perform as a first step.

### DIFF
--- a/deploy/azureDeployFunction.js
+++ b/deploy/azureDeployFunction.js
@@ -18,6 +18,7 @@ class AzureDeployFunction {
 
     this.hooks = {
       'deploy:function:deploy': () => BbPromise.bind(this)
+        .then(this.provider.initialise(this.serverless,this.options))
         .then(this.loginToAzure)
         .then(this.createFunction)
         .then(() => this.serverless.cli.log('Successfully uploaded Function'))
--- a/deploy/azureDeployFunction.js
+++ b/deploy/azureDeployFunction.js
@@ -18,7 +18,7 @@ class AzureDeployFunction {
 
     this.hooks = {
       'deploy:function:deploy': () => BbPromise.bind(this)
-        .then(this.provider.initialise(this.serverless,this.options))
+        .then(this.provider.initialize(this.serverless,this.options))
         .then(this.loginToAzure)
         .then(this.createFunction)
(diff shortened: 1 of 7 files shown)


aidanmorgan:
I am happy to re-work this based on any architectural guidance from the team, but these changes are needed to support proper variable substitution in provider fields, and I need them to make resource-group etc. changes.

provider/azureProvider.js
```diff
@@ -63,9 +63,16 @@ return constants.providerName;
...
 
   constructor (serverless) {
-    this.serverless = serverless;
     this.provider = this;
+    this.serverless = serverless;
+
     this.serverless.setProvider(constants.providerName, this);
+  }
+
+  initialise(serverless, options) {
```
pragnagopa: Rename to initialize

deploy/azureDeployFunction.js
```diff
@@ -18,6 +18,7 @@ class AzureDeployFunction {
 
     this.hooks = {
       'deploy:function:deploy': () => BbPromise.bind(this)
+        .then(this.provider.initialise(this.serverless,this.options))
```
pragnagopa: Is calling initialization before every command the only way to make sure serverless framework variables are properly initialized?

aidanmorgan: We need to do the initialise as late in the process as possible to ensure that the variable substitution has occurred. 

This provider seems to work differently, other providers appear to pass in all of the required parameters, rather than keeping them stored inside the provider like this one.

If there is a better place to put this I can change it - but this is as late in the process as I can find that ensures that the framework has configured everything we need.

provider/azureProvider.js
aidanmorgan: Sorry - I'm Australian :-)

aidanmorgan:
@pragnagopa I have changed initialise to initialize as requested, could you provide a suggestion about a more suitable way of initialising the provider?

pragnagopa:
Looks good. Will merge these changes. I will look into other providers to see if there is a better way to initialize.

PR #59, opened 2017-04-26
Adding an interactive login experience

lostintangent:
This PR introduces an interactive login experience (using the `az-login` component I'm working on), that allows devs to deploy/manage their Azure functions, without needing to explicitly create a service principal and then set the four env vars. This results in a simpler, less error-prone getting started experience. Additionally, this PR results in more code being removed, then added, which is always nice :)

With this change, if those env vars are set, then they will continue to be used (although it also supports the same env vars as Terraform as well, for increased interop). However, if they aren't set, then the CLI will initiate an interactive login by launching the user's default browser to the device login page (`https://aka.ms/devicelogin`), and copying the device code to the clipboard. Once login has been successfully completed, it will create a service principal behind the scenes, and persist in to disk, so that subsequent commands will naturally pick it up.

screen shot 2017-04-25 at 5 10 16 pm

As far as subscription selection goes, if the authenticated account only has a single subscription, it will simply select that one (preventing the user from needing to set a subscription ID env var). If there are multiple subscriptions, and the user is using the Azure CLI 2.0, it will look to see what their default subscription has been set to from there. If that doesn't exist, then it will fall back to prompting the user for the subscription they would like to use.

screen shot 2017-04-25 at 5 33 07 pm

This way, both the authentication and subscription selection process can be entirely guided.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,13 +21,14 @@
   ],
   "dependencies": {
     "async": "^2.1.2",
+    "az-login": "^0.1.4",
+    "azure-arm-resource": "1.6.1-preview",
     "bluebird": "^3.4.6",
     "chalk": "^1.1.3",
-    "lodash": "^4.16.6",
     "fs-extra": "^2.0.0",
-    "request": "2.79.0",
+    "lodash": "^4.16.6",
     "ms-rest-azure": "^1.15.4",
-    "azure-arm-resource": "1.6.1-preview",
+    "request": "2.79.0",
     "zip-folder": "1.0.0"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   ],
   "dependencies": {
     "async": "^2.1.2",
-    "az-login": "^0.1.5",
+    "az-login": "^0.1.6",
     "azure-arm-resource": "1.6.1-preview",
     "bluebird": "^3.4.6",
     "chalk": "^1.1.3",
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   ],
   "dependencies": {
     "async": "^2.1.2",
-    "az-login": "^0.1.4",
+    "az-login": "^0.1.5",
     "azure-arm-resource": "1.6.1-preview",
     "bluebird": "^3.4.6",
     "chalk": "^1.1.3",
--- a/provider/azureProvider.js
+++ b/provider/azureProvider.js
@@ -2,7 +2,6 @@
 
 const BbPromise = require('bluebird');
 const _ = require('lodash');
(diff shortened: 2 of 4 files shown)


fiveisprime:
Tested on Windows with serverless@1.12.0

- [x] [Deploy] Selected default subscription
- [x] [Deploy] Prompted for subscription after running `az account clear`
- [x] [Remove] Selected default subscription
- [x] [Remove] Prompted for subscription after clearing default

My double quote nitpick stands, but everything works as expected. :)

fiveisprime:
Update: I continued playing with this and, at one point, failed to auth saying that the code I pasted had expired. 🤷‍♀️

pragnagopa:
@lostintangent - Thanks for the PR.  Having support for interactive login will be a huge improvement from the current experience. Please submit PR against dev branch. 
Couple for questions:
- Is it a one time setup using the deviceCode?
- Is there a way user an specify name and password for ServicePricipal user created in the back ground? This will be helpful for customers as they can reuse service principal created.

@fiveisprime - Thanks for testing.

fiveisprime:
> Is it a one time setup using the deviceCode?

Yes, `az-login` stores the service principal in the user's home directory (in `~/.azure/azloginServicePrincipal.json`) and checks for that before initiating the interactive login.

I'm not sure what you mean for the other question though; `az-login` does attempt to use the existing environment variables so the current auth strategy (using the CLI then copying values to env vars) will continue to work. Maybe that answers your question?

---

What's the strategy with the dev branch? It's looking stale with the last merge almost a month ago. :/

pragnagopa:
dev is the current branch for making any updates. Once we are ready to release new npm package, dev will be merged to master.

fiveisprime:
bits in the latest publish match e6d4578; not the branch

Is the team using `git flow`? Is there a cadence for releases on this?

pragnagopa:
We are not using git flow. For now, we would like to keep master branch as "production" and keep using dev branch for making any updates. Interactive login is a huge improvement over the current experience.  We will release npm package once this change is tested and merged.

lostintangent:
Closing this in favor of submitting it to the dev branch.

cmatskas:
Hi all, could you explain the security implications of storing the SP credentials on disk or as environment variables. I love the idea of an interactive login but I'm concerned that this may open up some security wholes. An SP with Contributor rights is quite an elevated account to leave exposed. Thoughts?

lostintangent:
@cmatskas Just submitted an update to `az-login` that uses the native OS secrets store (e.g. keychain on macOS) to actually persist the SP secret. This way, the secret is never insecure at rest. Let me know what you think.

Also, let's collaborate on #63 which is the replacement of this PR. Thanks!

PR #61, opened 2017-04-26
Support service wide environment variables

### LINKED ISSUE #60: Unable to set service wide environment variables
The Serverless AWS provider supports service wide environment variables. See example at https://serverless.com/framework/docs/providers/aws/guide/serverless.yml/. Setting those with the Azure provider does not have any effect.

vjrantal:
Fixes #60.

### DIFF
--- a/provider/azureProvider.js
+++ b/provider/azureProvider.js
@@ -11,6 +11,7 @@ const https = require('https');
 const zipFolder = require('zip-folder');
 const request = require('request');
 const dns = require('dns');
+const jsonpath = require('jsonpath');
 const parseBindings = require('../shared/parseBindings');
 
 let resourceGroupName;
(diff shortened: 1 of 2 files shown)


vjrantal:
@christopheranderson @pragnagopa The approach on this one is that if during the deployment there are environment variables found from the `serverless.yml` the ARM template is patched "on the fly" to include those. The reason for this design decision is:

1. I don't know any other way to set environment variables from a command line that in the ARM template.
2. Template parameters were not used, because I don't think there is a syntax in ARM to loop through the parameters in the deeply nested JSON structure in the way required for this scenario.

I have tested that with this implementation variables from `serverless.yml` end up to the App Service Application settings. I also tested that changing the values and re-deploying does change them and removing variables removes them as expected.

provider/azureProvider.js
```diff
@@ -178,7 +179,24 @@ return new BbPromise((resolve, reject) => {
       }
     }
 
-    const template = JSON.parse(fs.readFileSync(templateFilePath, 'utf8'));
+    let template = JSON.parse(fs.readFileSync(templateFilePath, 'utf8'));
+
```
pragnagopa: Instead of injecting environment variables into ARM template, can you add them variables list? See above for how userParameters are passed into ARM template?

vjrantal: See point 2. at https://github.com/serverless/serverless-azure-functions/pull/61#issuecomment-297384088 for the explanation why I did not use parameters.

To elaborate, it would be easy to pass the environment variables as parameters (as you say), but I do not know a syntax in the ARM template JSON to loop through those parameters in the part of the JSON where that would be needed. If you do know this would be possible, please send a diff to the ARM template that shows how the looping of the parameters would be done under the list which is the value of the `appSettings` property.

pragnagopa: With the implementation you have, all the environment variables specified are injected into ARM template without an an option to explicitly choose. I would prefer customers providing any settings that need to be supplied to ARM template via parameters in yml file.
Can you try using variables:
- Specify AppSetitng in ARM template:  "[variables('KeyName')]": "[variables('EnvValue')]"
- Provide keyName in armTemplate variables
- If a value is not specified for the keyname, look up that keyName in serverless environment variables

vjrantal: The current implementation makes the environment variable configuration for the Azure provider behave like it behaves with the other Serverless providers. I think it is valuable to align the behavior between providers wherever possible. If you don't want to set some variable, then you don't define it in your `serverless.yml` file.

pragnagopa: Ok. Got it. Will go ahead and merge these changes.

Issue #8, opened 2017-02-20
Add "invoke function -log" feature


christopheranderson:
Add invoke function -log feature which will give us the log output for a given execution

pragnagopa:
Blocked on [issue](https://github.com/Azure/azure-webjobs-sdk-script/issues/1122)

lostintangent:
This would be immensely useful!

tbarlow12:
Will close this in favor of #373

PR #63, opened 2017-04-28
Adding an interactive login experience

lostintangent:
This PR introduces an interactive login experience (using the `az-login` component I'm working on), that allows devs to deploy/manage their Azure functions, without needing to explicitly create a service principal and then set the four env vars. This results in a simpler, less error-prone getting started experience. Additionally, this PR results in more code being removed, then added, which is always nice :)

With this change, if those env vars are set, then they will continue to be used (although it also supports the same env vars as Terraform as well, for increased interop). However, if they aren't set, then the CLI will initiate an interactive login by launching the user's default browser to the device login page (`https://aka.ms/devicelogin`), and copying the device code to the clipboard. Once login has been successfully completed, it will create a service principal behind the scenes, and persist in to disk, so that subsequent commands will naturally pick it up.

screen shot 2017-04-25 at 5 10 16 pm

As far as subscription selection goes, if the authenticated account only has a single subscription, it will simply select that one (preventing the user from needing to set a subscription ID env var). If there are multiple subscriptions, and the user is using the Azure CLI 2.0, it will look to see what their default subscription has been set to from there. If that doesn't exist, then it will fall back to prompting the user for the subscription they would like to use.

screen shot 2017-04-25 at 5 33 07 pm

This way, both the authentication and subscription selection process can be entirely guided.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,129 +1,75 @@
-# Azure Functions Plugin 
+# Azure Functions Serverless Plugin 
 
 This plugin enables Azure Functions support within the Serverless Framework.
 
 ## Getting started
 
-### 1. Get a Serverless Service with Azure as the Provider
+### Pre-requisites
 
-1. Recommend using Node v6.5.0
-1. Install the serverless tooling - `npm i -g serverless`
-1. Create boilerplate (change `my-app` to whatever you'd prefer): `serverless install --url https://github.com/azure/boilerplate-azurefunctions --name my-app`
-1. `cd my-app`
-2. `npm install`
+1. Node.js v6.5.0+ *(this is the runtime version supported by Azure Functions)*
+2. Serverless CLI. You can run `npm i -g serverless` if you don't already have it.
+3. An Azure account. If you don't already have one, you can sign up for a [free trial](https://azure.microsoft.com/en-us/free/) that includes $200 of free credit.
 
-### 2. Set up credentials
+### Create a new Azure service
 
-We'll set up an Azure Subscription and our service principal. You can learn more in the [credentials doc]( https://www.serverless.com/framework/docs/providers/azure/guide/credentials).
+1. Create a new service using the standard Node.js template: `serverless create -t azure-nodejs --name <APP_NAME>`
+2. CD into the generated app directory *(`cd <APP_NAME>`)*
+3. Install the app's NPM dependencies, which includes this plugin *(`npm install`)*
+4. Update the `service` property in your `serverless.yml` file, to ensure it specifies a unique app name:
 
-1. Set up an Azure Subscription
-
-    Sign up for a free account @ [https://azure.com](https://azure.microsoft.com/en-us/services/functions/).
-
-    Azure comes with a [free trial](https://azure.microsoft.com/en-us/free/) that includes $200 of free credit. 
-
-
-2. . Get the Azure CLI
-
-    ```
-    npm i -g azure-cli
+    ```yml
+    service: my-azure-functions-app # Name of the Azure function App you want to create
     ```
 
-3. Login to Azure
+### Deploy, test, and remove your service
 
-    ```
-    azure login
-    ```
-
-    This will give you a code and prompt you to visit [aka.ms/devicelogin](https://aka.ms/devicelogin). Provide the code and then login with your Azure identity (this may happen automatically if you're already logged in). You'll then be able to access your account via the CLI.
-
-4. Get your subcription and tenant id
-
-    ```
-    azure account show
-    ```
-
-    Save the subcription and tenant id for later
-
-5. Create a service principal for a given `<name>` and `<password>` and add contributor role.
-
-    ```
-    azure ad sp create -n <name> -p <password>
-    ```
-
-    This should return an object which has the `servicePrincipalNames` property on it and an ObjectId. Save the Object Id and one of the names in the array and the password you provided for later. If you need to look up your service principal later, you can use `azure ad sp -c <name>` where `<name>` is the name provided originally. Note that the `<name>` you provided is not the name you'll provide later, it is a name in the `servicePrincipalNames` array.
-
-    Then grant the SP contributor access with the ObjectId
+1. Deploy your new service to Azure! The first time you do this, you will be asked to authenticate with your Azure account, so the `serverless` CLI can manage Functions on your behalf. Simply follow the provided instructions, and the deployment will continue as soon as the authentication process is completed.
 
     ```bash
-    azure role assignment create --objectId <objectIDFromCreateStep> -o Contributor
+    serverless deploy
     ```
 
-6. Set up environment variables
+    > Note: Once you've authenticated, a new Azure "service principal" will be created, and used for subsequent deployments. This prevents you from needing to manually login again. See [below](#advanced-authentication) if you'd prefer to use a custom service principal instead.
 
-     You need to set up environment variables for your subscription id, tenant id, service principal name, and password. 
+2. Invoke a function, in order to test that it works:
 
     ```bash
-    # bash
-    export azureSubId='<subscriptionId>'
-    export azureServicePrincipalTenantId='<tenantId>'
-    export azureServicePrincipalClientId='<servicePrincipalName>'
-    export azureServicePrincipalPassword='<password>'
+    serverless invoke -f hello
     ```
 
-    ```powershell
-    # PowerShell
-    $env:azureSubId='<subscriptionId>'
-    $env:azureServicePrincipalTenantId='<tenantId>'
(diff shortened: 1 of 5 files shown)


lostintangent:
@pragnagopa @fiveisprime I did a bunch more testing, and everything looks good. Let me know what you think!

provider/azureProvider.js
```diff
@@ -112,16 +92,14 @@ return this.parsedBindings;
...
-        if (error) {
-          reject(error);
-        } else {
-          principalCredentials = credentials;
-          resolve(credentials);
-        }
-      });
+  Login() {
+    return login({ interactiveLoginHandler: (code) => {
+        this.serverless.cli.log(`Paste this code (copied to your clipboard) into the launched browser, and complete the authentication process: ${code}`);
```
pragnagopa: Can you add more explicit logging to indicate that this is going to create a service principal behind the scenes as new Azure customers will not be familiar with SP user.

fiveisprime:
I can't get this installed. I get an error that the az-login version isn't available.

I was testing via `$ npm i lostintangent/serverless-azure-functions` 

Did you miss publishing `az-login@0.1.7`?

lostintangent:
@fiveisprime Apologies! Published. Try again please :)

provider/azureProvider.js
lostintangent: I agree that new Azure customers wouldn't be familiar with SPs, which is why this PR doesn't currently log out that it's creating one. The auto-generated SP is meant to be an implementation detail that simply provides the ability to login interactively once, and then continue to use the CLI without needing to login again. I'd personally vote not to add any logging here, because it might just complicate the experience.

provider/azureProvider.js
```diff
@@ -75,21 +65,11 @@ return constants.providerName;
     this.options = options;
 
     return new BbPromise((resolve, reject) => {
-        subscriptionId = this.getSetting(azureCredentials.azureSubId);
-        servicePrincipalTenantId = this.getSetting(azureCredentials.azureServicePrincipalTenantId);
```
pragnagopa: Remove getSetting function as well

provider/azureProvider.js
pragnagopa: Does ServicePrincipal user created here show up in [portal](https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-create-service-principal-portal)? If yes, I think we should let users know about creating SP user.

provider/azureProvider.js
lostintangent: Done!

provider/azureProvider.js
```diff
@@ -102,27 +82,14 @@ return constants.providerName;
...
-      if (process.env.hasOwnProperty(k)) {
-        if (k.toLowerCase() === key.toLowerCase()) {
-      return process.env[k];
-        }
-      }
-    }
-  }
+  Login() {
+    return login({ interactiveLoginHandler: (code) => {
+        this.serverless.cli.log(`Paste this code (copied to your clipboard) into the launched browser, and complete the authentication process: ${code}`);
```
pragnagopa: Does SP user created here show up in [portal](https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-create-service-principal-portal)? If yes, I think we should log indicating SP user being created.

lostintangent: That implies users are going to go into the portal and explore their `Azure Active Directory` settings (is that really that common for new Azure users?) Can't we just keep the logging simple and expand it based on feedback? If we log super verbose messages, we run the risk of making it overwhelming/overcomplicated for everyone, as opposed to accommodating the few that would care to know a SP is being created.

I actually think the current logging for the `deploy` is already way too verbose as it is.

pragnagopa:
@lostintangent - Please update readme as well. Remove all the info that about setting up environment variables.

provider/azureProvider.js
pragnagopa: As long as it is mentioned somewhere that SP user is being created automatically, I am ok with not logging. Mentioning it in ReadMe is works!

pragnagopa: Logs can updated to 'verbose' so that more logs show up only when looking for verbose logs. This will default logs simpler.

lostintangent:
@pragnagopa Just updated the `README`, let me know what you think. It's quite a bit simpler :)

I'll update the Serverless docs after this PR is merged.

README.md
```diff
@@ -1,129 +1,75 @@
...
-### 1. Get a Serverless Service with Azure as the Provider
+### Pre-requisites
 
-1. Recommend using Node v6.5.0
-1. Install the serverless tooling - `npm i -g serverless`
-1. Create boilerplate (change `my-app` to whatever you'd prefer): `serverless install --url https://github.com/azure/boilerplate-azurefunctions --name my-app`
-1. `cd my-app`
-2. `npm install`
+1. Node.js v6.5.0+ *(this is the runtime version supported by Azure Functions)*
+2. Serverless CLI. You can run `npm i -g serverless` if you don't already have it.
```
pragnagopa: serverless-azure-functions needs to be installed before running command: serverless create -t azure-nodejs --name <APP_NAME>

README.md
```diff
@@ -1,129 +1,75 @@
...
+### Deploy, test, and diagnose your Azure service
 
-5. Create a service principal for a given `<name>` and `<password>` and add contributor role.
+1. Deploy your new service to Azure! The first time you do this, you will be asked to authenticate with your Azure account, so the `serverless` CLI can manage Functions on your behalf. Simply follow the provided instructions, and the deployment will continue as soon as the authentication process is completed.
 
-    ```
-    azure ad sp create -n <name> -p <password>
+    ```bash
+    serverless deploy
     ```
```
pragnagopa: Add command to deploy one function : serverless deploy function -f <functionName>

README.md
lostintangent: This "create" support is part of the core serverless CLI. Also, the "Serverless-Azure-Functions" module is listed as an NPM dependency of the generated app, so running "npm install" will install it.

pragnagopa: I just tried this and it fails with error:  Cannot find module 'serverless-azure-functions'

README.md
```diff
@@ -1,129 +1,75 @@
...
 
     ```bash
-    azure role assignment create --objectId <objectIDFromCreateStep> -o Contributor
+    serverless invoke -f hello
     ```
 
-6. Set up environment variables
-
-     You need to set up environment variables for your subscription id, tenant id, service principal name, and password. 
+3. View the output logs of your newly deployed function:
```
pragnagopa: Change to : Stream the Function Logs. We have an [issue ](https://github.com/serverless/serverless-azure-functions/issues/8)tracking fetching logs for an invocation.

README.md
```diff
@@ -1,129 +1,75 @@
...
+1. Node.js v6.5.0+ *(this is the runtime version supported by Azure Functions)*
+2. Serverless CLI. You can run `npm i -g serverless` if you don't already have it.
+3. An Azure account. If you don't already have one, you can sign up for a [free trial](https://azure.microsoft.com/en-us/free/) that includes $200 of free credit.
 
-### 2. Set up credentials
+### Create a new Azure service
 
-We'll set up an Azure Subscription and our service principal. You can learn more in the [credentials doc]( https://www.serverless.com/framework/docs/providers/azure/guide/credentials).
+1. Create a new service using the standard Node.js template: `serverless create -t azure-nodejs --name <APP_NAME>`
+2. CD into the generated app directory: `cd <APP_NAME>`
```
pragnagopa: creating a template is not generating App_Name directory. Remove this for now. This should be fixed in core serverless CLI

lostintangent: Actually, I just need to change this sample to use `-p` instead, and it will create the directory as expected. This was a doc bug.

README.md
lostintangent: @pragnagopa What version of the Serverless CLI do you have installed? It requires `1.9.0`+, which is when the template was released. I just tried it again and it worked for me.

lostintangent:
@pragnagopa I addressed all of your feedback, and re-tested the `serverless create -t azure-nodejs` and can't repro the issue you mentioned. Let me know if you're using Serverless CLI 1.9.0+ and seeing this problem, and I'll look into it again.

mamaso:
There may be issues with multiple subscriptions:

image

pragnagopa:
@mamaso - this PR is merged now. Can you open an issue with repro steps to track issue with  multiple subscriptions?

lostintangent:
@mamaso Could you check whether you have the `azureSubId` or `ARM_SUBSCRIPTION_ID` env variable set? And if so, could you validate that it is set to an ID of a subscription that is associated with the identity you authenticated with?

PR #65, opened 2017-04-30
Adding App Insights support

lostintangent:
This PR updates the existing ARM deployment templates, to create an App Insights instance, and bind the functions to it. This way, devs can deploy their functions, invoke them, etc. and then begin to monitor them without needing to do anything extra to set this up. Since App Insights provides a pretty permissive free-tier (1 GB data/month), it seems reasonable to create this resource by default.

Some questions to discuss around the design of this feature:

- [ ] Do we want to allow easily disabling provisioning an AI instance, or is it acceptable for now to just auto-create it? The user always has the option to provide a custom ARM template as a fallback, though that may be a little heavy-handed for this scenario.

- [x] Is it safe to always enable the beta runtime, or do we need to toggle this conditionally?

- [ ] AI is only available in four regions currently, so we can't reliably align the AI location with the specified region in the `serverless.yml` file. Is it sufficient for now to just lock it to `eastus`? Since this feature is trying to be a transparent value add, I'd hate to error out if a user tried to deploy a function that West US, and the AI instance couldn't be assigned to that location (for example).

### DIFF
--- a/provider/armTemplates/azuredeploy.json
+++ b/provider/armTemplates/azuredeploy.json
@@ -89,7 +89,7 @@
                         },
                         {
                             "name": "FUNCTIONS_EXTENSION_VERSION",
-                            "value": "beta"
+                            "value": "~1"
                         },
                         {
                             "name": "WEBSITE_CONTENTAZUREFILECONNECTIONSTRING",
--- a/provider/armTemplates/azuredeploy.json
+++ b/provider/armTemplates/azuredeploy.json
@@ -25,10 +25,21 @@
     "variables": {
         "functionAppName": "[parameters('functionAppName')]",
         "hostingPlanName": "[parameters('functionAppName')]",
+        "insightsName": "[concat(parameters('functionAppName'), '-insights')]",
         "storageAccountName": "[concat(uniquestring(resourceGroup().id), 'azfunctions')]",
         "storageAccountid": "[concat(resourceGroup().id,'/providers/','Microsoft.Storage/storageAccounts/', variables('storageAccountName'))]"
     },
     "resources": [
+        {
+            "type": "microsoft.insights/components",
+            "kind": "Node.JS",
+            "name": "[variables('insightsName')]",
+            "apiVersion": "2014-04-01",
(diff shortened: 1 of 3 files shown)


fiveisprime:
I'd like some feedback on what else the beta flag changes.

lostintangent:
@fiveisprime Yep, agreed! I decided to open this PR to have a discussion around this, but we may need to add some extra config options to make this as unobtrusive as possible, while still lighting up an opinionated/awesome experience out of the box.

cmatskas:
###Beta 
At the moment the ~beta flag is to provide AI support. Unfortunately, since this is also the way the team tests new features, it could break things in the interim. I think AI will go mainstream near or around /Build, so not long to go. 

###AI Region
Region wise, yes, we could default to EastUS and then let users change the AI instrumentation key to the one they want to point to.

###Auto-provisioning
I would say that since this is a free service, there's no harm adding it by default. I would also like to avoid having a too complex `.yml` configuration file. The less options the better

pragnagopa:
@christopheranderson - Are we OK with changing default ARM template to opt-in to beta?

cmatskas:
@prgnagora & @christopheranderson - I think there's an issue with the current implementation of the ~beta code and I'll quote it below:

### Known issues

- Dependencies don’t show up automatically
- Cold start times are really bad (this is because we’ve not optimized this build, so it’s got very slow cold start and some runtime perf issues)

Linked from here: https://github.com/Azure/Azure-Functions/wiki/App-Insights-Early-Preview

fabiocav:
@pragnagopa @cmatskas  the AI work has been merged into the main branch and is part of the current release (1.0.10917), so the `beta` version will no longer be required (and will likely be removed and/or used for other features in the future)

lostintangent:
@fabiocav Good to know! I'll remove setting that explicitly in this PR then.

lostintangent:
FYI, I updated the PR to not explicitly set the beta version. I tested the E2E and it still works great, so as @fabiocav mentioned, that doesn't appear to be necessary anymore.

Issue #71, opened 2017-06-06
Deployment ignores errors from the Kudu command API


vjrantal:
If there are errors returned from the Kudu command API during deployment, they are silently ignored.

To illustrate what I mean, I applied the following diff to the code:

```
diff --git a/provider/azureProvider.js b/provider/azureProvider.js
index 4c3c4d4..64b67cc 100644
--- a/provider/azureProvider.js
+++ b/provider/azureProvider.js
@@ -542,6 +542,8 @@ return new BbPromise((resolve, reject) => {
     };
 return new BbPromise((resolve, reject) => {
       request(options, (err, res, body) => {
+        console.log('Received status: ' + res.statusCode);
+        console.log(body);
         if (err) {
           reject(err);
         }
```

After that, a certain deploy command printed output like this:

```
Serverless: Syncing Triggers....Response statuscode: 200
Serverless: Running Kudu command del package.json...
Received status: 200
{ Output: '',
  Error: 'Could Not Find D:\\home\\site\\wwwroot\\package.json\r\n',
  ExitCode: 0 }
Serverless: Uploading pacakge.json ...
Serverless: Running Kudu command npm install --production...
Received status: 500
<html><head><title>500 - The request timed out.</title></head><body>  <font color ="#aa0000">         <h2>500 - The request timed out.</h2></font>  The web server failed to respond within the specified time.</body></html>
Serverless: Successfully created Function App
```

In above, observe how in the first call, the status is 200, but error is seen from the content of the body and in the second case, the status is 500. In both cases, the errors are ignored and eventually `Successfully created Function App` is printed even though there were errors.

Fix proposal is to look at both status code and body content in all Kudu commands and reject the related Promise in case errors are identified.

ps. On purpose, I am not providing reproduce steps since I intended this one to be only about proper error handling in case any type of errors occur.

fiveisprime:
Thank you so much for the detailed report! 🎉 

I'll get some additional error handling added and deployed this week (hopefully today).

fiveisprime:
@vjrantal Can you deploy your test/broken function with my fork?

```sh
$ npm install fiveisprime/serverless-azure-functions
```

See if it errors out correctly. I made some changes to that particular piece, but I'm still working through all of the other places where this could come up.

lostintangent:
@fiveisprime Did #72 end up resolving this issue?

lostintangent:
Closing this as resolved by #72. Please reopen if you experience any further issues. Thanks!

Issue #68, opened 2017-05-06
Cannot read property 'tokenCache' of undefined


yhersh:
Hello,

I have followed the [azure quickstart guide](https://serverless.com/framework/docs/providers/azure/guide/quickstart/) and when running the the first `serverless deploy` command I get the following:

`Serverless: Packaging service...
Serverless: Logging in to Azure

  Type Error ---------------------------------------------

     Cannot read property 'tokenCache' of undefined

     For debugging logs, run again after setting the "SLS_DEBUG=*" environment variable.

  Stack Trace --------------------------------------------

TypeError: Cannot read property 'tokenCache' of undefined
    at authenticate.then (/Users/yaron.hersh/work/guyg/apriltex-azfunc-dblist/node_modules/az-login/index.js:209:40)
    at process._tickDomainCallback (internal/process/next_tick.js:135:7)
From previous event:
    at PluginManager.invoke (/usr/local/lib/node_modules/serverless/lib/classes/PluginManager.js:234:22)
    at PluginManager.run (/usr/local/lib/node_modules/serverless/lib/classes/PluginManager.js:253:17)
    at Serverless.run (/usr/local/lib/node_modules/serverless/lib/Serverless.js:97:31)
    at serverless.init.then (/usr/local/lib/node_modules/serverless/bin/serverless:23:50)`

My enviroment setup is the as the guide suggested.

`serverless --version` : 1.12.1
`node`: v6.10.2

Thanks

lostintangent:
@yhersh Thanks for filing this bug report! I was able to repro this issue by setting an invalid value for one of the four env vars outlined in the getting started guide. I'm not sure whether that is exactly what you're experiencing, but it's definitely a bug that I'll fix ASAP.

Could you double-check that your subscription ID, client ID, client secret and tenant ID env vars are all set correctly? Thanks!

lostintangent:
@yhersh I fixed the issue I mentioned previously and published an update. That said, that fix will simply make the error more friendly if the env vars you set contain invalid credentials (e.g. there's a typo in the client ID, the service principal is expired), so if that isn't the issue, then I'll need to investigate it further.

sandeep-paliwal:
I am also having similar issue although I have set env variables correctly.

> TypeError: Cannot read property 'tokenCache' of undefined
    at authenticate.then (/Users/spaliwal/work/performance/serverless-performance/serverless-performance/node_modules/az-login/index.js:209:40)
[...]

Issue #28, opened 2017-02-24
Throw errors if the Rest API calls return non success HTTP status code


pragnagopa:
Throw errors if the Rest API call return 4xx or 5xx  in [azureProvider](https://github.com/serverless/serverless-azure-functions/blob/master/provider/azureProvider.js)

lostintangent:
Closing this as a dupe of #71

PR #72, opened 2017-06-07
Add lint, simplify some code, add error handling.

### LINKED ISSUE #6: Throw helpful errors on bad creds
Currently, we just leak the errors from the aad libraries if the credentials aren't provided correctly. We should catch the environment variables set incorrectly and throw errors. We should help provide meaningful help if the credentials are set up correctly, but fail.

fiveisprime:
Alright, this looks intimidating, but I swear it's not. :)

To test this, create a serverless app and install the plugin from my fork via
`$ npm install fiveisprime/serverless-azure-functions` then run through the
various commands. You may run into DNS issues when removing/recreating,
which is already a tracked issue (it's much further down the stack yo).

Updated to the latest version of az-login to fix the error handling for failed
login attempts.

Added ESLint test command, cleaned up all existing lint issues, and cleaned up
some confusing/unused code. Normalized the HTTP request usage so they all use the
request module rather than some using the https module.

run `$ npm test` to run the lint (please do this before committing).

Changed the invoke command to no longer require logging in to Azure. I left the
code in place to facilitate future logging based on invocation ID (tracked at
azure-webjobs-sdk-script/issues/1122).

Updated the log streaming code to use a stream rather than concating strings. The
error handling and reconnect logic wasn't actually working in the existing code,
this all works as intended now and the output is cleaner (no more line breaks
between logs).

Moved all "constants" into a separate config to facilitate some testing later on.
Adding tests is on my list, but will have to come a bit later (sorry) once we can
agree on a framework.

_Note: These changes are proposed against master. I've spoken with Pragna and
received approval to change the branching strategy, which I'll document in the
wiki._ <3

Closes #6, #28, #71, #68

### DIFF
--- a/provider/azureProvider.js
+++ b/provider/azureProvider.js
@@ -6,15 +6,15 @@ const resourceManagement = require('azure-arm-resource');
 const path = require('path');
 const fs = require('fs');
 const fse = require('fs-extra');
-const https = require('https');
 const zipFolder = require('zip-folder');
 const request = require('request');
 const dns = require('dns');
 const jsonpath = require('jsonpath');
 const parseBindings = require('../shared/parseBindings');
 const { login } = require('az-login');
+const config = require('../config');
 
-let pkg = require('../package.json');
+const pkg = require('../package.json');
 
 let resourceGroupName;
 let deploymentName;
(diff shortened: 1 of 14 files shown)


provider/azureProvider.js
```diff
@@ -227,64 +199,58 @@ class AzureProvider {
...
-      'host': functionAppName + constants.scmDomain,
-      'port': 443,
-      'path': constants.masterKeyApiApth,
-      'headers': {
-        'Authorization': constants.bearer + principalCredentials.tokenCache._entries[0].accessToken,
-        'Content-Type': constants.jsonContentType
+      url: `https://${functionAppName}${config.scmDomain}${config.masterKeyApiPath}`,
+      json: true,
+      headers: {
+        Authorization: config.bearer + principalCredentials.tokenCache._entries[0].accessToken
```
lostintangent: As a future optimization, it might clean things up to use the modified `request` instance that `az-login` provides. That way we don't need to construct these authenticated `request` instances "manually".

fiveisprime: Yes! I really really really considered doing that in this set. It's on the list and thank you for bringing it up. 👍

lostintangent:
LGTM!

provider/azureProvider.js
pragnagopa: 👍

pragnagopa:
Looks good 👍

PR #79, opened 2017-07-09
Deploy azure function apps more like in aws/gce

mabels:
Hi,

i had trouble to deploy a project containing multiple shared resources on azure using your nice serverless-azure-function plugin. 
I don't want to use webpack to get it work. 
And as my application supports multiple function/lambda providers like aws/gce. I want to have a more or less the same looking process and filesystem structure.
So I modified your plugin to get it work. Now I use the serverless provided packaging and add the entry point handlers to the serverless packaged zip file. 
During the deployment I now copy over the pre-packaged zip from the serverless framework, thats all.
There is no need to run npm install on the azure systems.
I would like to know what you are thinking, is the way to go with your plugin?
If yes i would cleanup my current changes for a later integration. 
This changes is only in a proove of concept status.

thx in advance  
Here is my test project: https://github.com/mabels/talks/tree/master/is-this-simple/src/azure


DDeme:
I have similar issues. I can put a afford to it, Could somebody provide me a info what needs to done to merge this PR.

marsher:
+1

lostintangent:
Sorry for the late response here! #83 appears to solve a similar purpose, but leveraging more of the Serverless packaging system (which helps clean this plugin up a little more). 

@mabels What are your thoughts on reconciling these two PRs? Does #83 look like it addresses your requirements?

lostintangent:
Closing this PR in favor of #83, which addresses the same feedback. Please let me know if there are still improvements we should make and/or send any follow up PRs. Thanks again for the contribution!

PR #82, opened 2017-08-07
Fix typo in console output when uploading package.json

elmundio87:
Was: `Serverless: Uploading pacakge.json...`
Now: `Serverless: Uploading package.json...`


lostintangent:
Thanks for catching this!

PR #83, opened 2017-08-13
improve packaging for azure functions

### LINKED ISSUE #34: Deployed directory structure and dependencies
What is the proper behavior in regards to dependencies? Currently, only the entry file is uploaded to Azure and a `function.json` is automatically created and uploaded to the function directory.  

i.e.

|-handler.js
|-serverless.yml

becomes,

|-functionName/
|--index.js
|--function.json

However if I create a new directory and file which is required by the entry file (handler.js in the example) the dependencies do not get uploaded.

i.e.

|-lib/
|--other.js     <- library file
|-handler.js   <- requires library file
|-serverless.yml

still produces,

|-functionName/
|--index.js
|--function.json

Running the function will throw a module not found error when required. Meaning the dependencies were not bundled with the file.

This differs from the behavior of other providers. Particularly AWS where all the dependencies are automatically bundled and uploaded.

Is this behavior expected or am I missing something? Will there be changes made to function similarly to the other providers?

Thanks!

Example Repo:
1. `git clone https://github.com/HyveMynd/azure-test.git`
2. `npm i`
3. `sls deploy`

### LINKED ISSUE #69: npm install on deploy
Uploading project `package.json` to Azure and then running `npm install` is not a good idea. It might be easy for quick start development, but should be avoided in any production-ready project. [NPM install is not immutable](https://ponyfoo.com/articles/immutable-npm-dependencies). So package we deployed to 'test' environment and the one we deployed to production environment might not be the same. 

This plugin should at least support a custom flag like `-no-install` or something similar that would prevent running `npm install` in Azure.

Trane9991:
Hello,
In this PR:

- Generating `function.json` file per function, with name `<func>-function.json` as the pre-package hook so that it can be packaged with Serverless native packaging process.

- Introduce `scriptFile` key for the `function.json`, so handler file doesn't need to be named `index.js`

- Renaming `<func>-function.json` to the `function.json` with *Kudo Command* after the deployment.

- Removed unneeded "zip-folder" dependency and its child dependencies (archiver, etc)

- Works with individual packaging option

- Works for "sls deploy" and "sls deploy function -f <func>"

Fixes #34 
Fixes #69

A bit duplicates PR #79


lostintangent:
@Trane9991 Thanks for this contribution! Do you think this PR supersedes #79? If not, could you speak a little bit about where there is duplication? 

I'll do a through review and testing later this evening. Thanks again!

Trane9991:
Hey @lostintangent,
Yes, it supersedes it.
My PR is solving the same problem, but a little bit different and more complete way.

Benefits of my solution versus #79:
- Completely takes advantage of Serverless packaging system, which helps to reduce the number of dependencies (`fse` and a lot of libs for working with zip files)
- Support of both deployment commands: `sls deploy` and `sls deploy function`
- Support of "individual packaging"
- Simpler and cleaner code

lostintangent:
@Trane9991 Awesome, thanks for confirming that.

provider/azureProvider.js
```diff
@@ -567,49 +564,48 @@ class AzureProvider {
...
         }
-      });
+      };
+
+      fs.createReadStream(functionZipFile)
+        .pipe(request.put(options, (uploadZipErr, uploadZipResponse) => {
+          if (uploadZipErr) {
+            reject(uploadZipErr);
+          } else {
+            // fse.removeSync(functionZipFile);
```
lostintangent: Can this line be removed? Now that the Serverless Framework handles creating the zip file, do they also clean it up?

deploy/lib/createFunctions.js
```diff
@@ -1,31 +1,19 @@
...
     });
 
-    const packageJsonFilePath = path.join(this.serverless.config.servicePath, 'package.json');
-    let createFunctionsPromise = BbPromise.all(createFunctionPromises)
-                                .then(() => this.provider.syncTriggers())
-                                .then(() => this.provider.runKuduCommand('del package.json'));
-
-    if (fs.existsSync(packageJsonFilePath)) {
-      return createFunctionsPromise.then(() => this.provider.uploadPackageJson())
-            .then(() => this.provider.runKuduCommand('npm install --production'));
```
lostintangent: This is probably a stupid question, but after this PR, is an NPM install ever automatically run on behalf of the developer as part of either the package/deploy commands? If not, do we need to validate that the function's dependencies have been installed?

deploy/azureDeploy.js
```diff
@@ -14,13 +15,18 @@ class AzureDeploy {
...
       this,
+      createEventsBindings,
       loginToAzure,
       cleanUpFunctions,
       CreateResourceGroupAndFunctionApp,
       createFunctions
     );
 
     this.hooks = {
+      'before:package:initialize': () => BbPromise.bind(this)
```
lostintangent: Should we be handling this logic as part of a new package command implementation? As opposed to handling package events in the deploy plugin? It probably doesn't matter too much, but it may be arguably cleaner to separate the package and deploy steps.

lostintangent: Since this PR allows calling the "serverless package" command directly, it would definitely seem to make more sense to split that implementation from the deploy command.

Trane9991: @lostintangent this PR doesn't allow you to call `sls package` explicitly, it just uses default one from Serverless. PR #79 has `sls package` implemented. We could implement it if needed.

deploy/lib/createFunctions.js
Trane9991: Nope, `npm install` will never run automatically. But we could think of adding some flag, so the user can explicitly run it if he wants to.

Not sure if we need to validate anything this is developers responsibility, aws provider works that way. You have only dependencies what you packed.

provider/azureProvider.js
Trane9991: Current behavior doesn't clean up the zip file but replacing it on each run.

It's useful for debugging, so you can unpack it and verify that all things that you want were packaged.

lostintangent: That makes sense. Should we delete this comment then?

Trane9991: @lostintangent yep, I will do it a bit later

deploy/azureDeploy.js
lostintangent: My initial thought is that it would be great to implement the `sls package` command as part of this PR. That way, we're enabling folks to use `package` and `deploy` as needed, and have things behave as they would expect. Additionally, that seems like it would allow this PR to fully supersede #79.

Trane9991: @lostintangent ok, I will implement it soon :)

Trane9991: @lostintangent `sls package` works now

Trane9991:
@lostintangent are we ready to merge it?

kunjan13:
is there a work around to include custom libraries/modules of nodejs while deploying azure functions?

Right now only functions.json and index.js gets deployed.

Trane9991:
This PR going to allow to include *node_modules* or any other directories.

lostintangent:
Yep! This looks great to me. Thanks again for this amazing contribution. We'll need to update the docs to include mention of the new "package" support, which I'll tackle this week.

Let me know if you have other feedback/ideas in the future!

hoffmannjan:
`Error: Cannot find module './shared/db'`

Still got this error, even after 

```
package:
  include:
    - shared/**
    - shared/db.js
```

drandx:
@lostintangent This feature hasn't been released yet right?. If it is, is there any new configuration on the serverless.yml needed in order to deploy the dependencies to Azure. Also, thanks @Trane9991 for the contribution!

PR #84, opened 2017-08-26
fix "invoke" command, by uncommenting "login" steps

Trane9991:


lostintangent:
Thanks for this change! You're right, that without this change, trying to invoke non-anonymous functions fails unfortunately. Could we check the `authLevel` of the function being invoked, and skip authentication if it's `anonymous`? That might be nice to simplify/speed up this operation for anonymous functions.

Additionally, I just tested this change, and it doesn't appear to work as-is with an HTTP-triggered function. I was able to get it working by adding the following HTTP request options on line 387 in `azureProvider.js`.

```javascript
headers: {
  'x-functions-key': functionsAdminKey
}
```

Did this work for you? Maybe I did something wrong and you can disregard this comment :)

Trane9991:
@lostintangent you are right. The code is responsible for invoking HTTP trigger lacking this header.

lostintangent:
Closing due to inactivity. I'll send out a PR with this change + the fix I mentioned.

Issue #93, opened 2017-11-19
trouble packaging/uploading local files and directories


huemansdesign:
Hi all, maybe Im not doing something correct. It seems like regardless of how I approach packaging local files/modules only my handler gets uploaded to Azure. I have to manually upload zip via Debug Console. I've tried packaging individually as well. Can you please tell me if I'm doing something wrong?

package:
  individually: true  
  exclude: 
    - libs/**
    - .bak/**

functions: 
  DService:
      handler: index.handler
      package:
        include:
          - config/**
          - DQuery.js 
          - index.js
          - modules/**
      events: 
        - http: true
          x-azure-settings:

yaron-idan:
We also encountered this issue, and it seems to be solved in this PR - https://github.com/serverless/serverless-azure-functions/pull/83. Perhaps the maintainers should release a new version that will contain this changes.

KosnIre:
Okay, so I wound up here because I had the exact problem in #34, but since this is the open ticket, I'll be describing my solution here. Sorry for the length.

At the time of writing, the most recent release is `serverless-azure-functions 0.4.0` and @yaron-idan is correct that #83 is **not** currently in a released version (it's just in `master`). So, here's what I did:

## Summary
1. `npm i --save git+ssh://git@github.com/serverless/serverless-azure-functions.git#b248b5bb514ad5e96d94204c957772171122a0b4`
2. `sls remove` and `sls deploy` (might not be necessary, but gets you a clean App)
3. a) Manually upload any `<yourFunction>-function.json` file via [https://<your-app-name>.scm.azurewebsites.net/dev](https://<your-app-name>.scm.azurewebsites.net/dev) in the corresponding function directory.
  b) Rename the uploaded file to `function.json` (you can right-click the file or hit F2)

## In-Depth

A couple of notes about my environment. I'm running `node v6.12.0` and `sls
My plugins look like this:
```
plugins:
  - serverless-webpack
  - serverless-offline
  - serverless-azure-functions
```
I'm also running `babel` and at the moment, both babel and webpack seem to be working properly.

1. `npm i --save git+ssh://git@github.com/serverless/serverless-azure-functions.git#b248b5bb514ad5e96d94204c957772171122a0b4`
That is the changeset that #83 got merged into `master`.

2. I also did a full `sls remove` and `sls deploy` here because Azure was giving me some errors (this might not be necessary for some people).

[...]

Issue #10, opened 2017-02-20
Add "webpack" feature


christopheranderson:
Should add a feature which webpacks the contents before sending them to Azure for better cold start performance.

This is blocked on the latest Feb runtime going out.

hoffmannjan:
anybody use serverless-webpack with serverless-azure-functions?

I was trying, but all I got deployed is bundled hander.js, without function.json so in azure-portal functions isn't visible.

HyperBrain:
If someone checked Azure with `serverless-webpack`, please put the results in https://github.com/serverless-heaven/serverless-webpack/issues/128. I'll try to gather all working integrations and compatibility issues there, so that we can make support of the plugin possible with all providers. Until now the OpenWhisk plugin has been fully tested and is added to the compatibility table.
It would be great to have the Azure plugin in there too - of course with a fully working integration 😄

HyperBrain:
@lostintangent Any news here? It would be great to test the provider plugin working with `serverless-webpack` and get it working properly. See my comment above.

KosnIre:
Hey @HyperBrain I have been toying with `serverless-webpack` combined with `serverless-azure-functions`. It seems this is very important for Azure Functions because it's slow to load many small files (e.g. the npm_modules dir).

I was able to get the combo running successfully, but there is a small problem. You have to manually upload the `function.json` files to Azure. See [my writeup here](https://github.com/serverless/serverless-azure-functions/issues/93#issuecomment-350599037).

HyperBrain:
Hi @KosnIre . That sounds good. 

Regarding the packaging of the JSON files in each package, you could use the `slsw.lib.serverless` object to add the function json files to each of the packages. If it would not be critical, that you add every json to each zip, you can fetch all function names easily in your webpack config:

```js
// webpack.conf.js
const slsw = require('serverless-webpack');

// Get the names of all functions from the current serverless instance
const allFuncNames = slsw.lib.serverless.service.getAllFunctions();

// Get the complete function objects from the serverless service.
const funcObjects = allFuncNames.map(allFuncNames, name => {
  return slsw.lib.serverless.service.getFuction(name);
});

// Now use the func vars above (or anything else retrieved from the serverless instance)
// to build the parameters for the webpack copy plugin to add all files you need.
```

As you have access to the complete serverless instance, you can use it to
retrieve anything from there - exactly as any serverless plugin would do.

KosnIre:
[...]

PR #105, opened 2018-01-10
Add <name>-function.json files to webpack package

KosnIre:
## Summary
Add functionality to include all `<name>-function.json` files in the .zip packages when using `serverless-webpack` with `serverless-azure-functions`.

## Status
_READY_

## Changes
- Added `before:webpack:package:packageModules` hook to `azurePackage.js`
- Added new file `webpackFunctionJson.js` to handle behavior related to moving 1 or more `<name>-function.json` files into the .webpack folder. This must occur _after_ the webpack folder is created, and _before_ the files are zipped. Once the files are moved, `serverless-webpack` handles the zip and the upload works as usual.
- Changed `azurePackageFunction.js` hook to `before:deploy:function:packageFunction`. This helps maintain consistent order of operations between `sls deploy` and `sls deploy function`.
- Added `serverless-webpack` (and `webpack` as peer-dependency of Serverless Webpack) to `optionalDependencies` in `package.json`.

## Testing
I have created [a sample repository](https://github.com/KosnIre/serverless-azure-webpack-sample) (also WIP) to test that the implementation is working as expected.

## Issues Addressed
#10 
#93


KosnIre:
@HyperBrain This update is working nicely during a `sls deploy`, but I hit a snag with `sls deploy function` and I could use an expert opinion. Any other knowledgeable person is welcome to chime in.

When I run sls deploy function, things are happening in a different order than sls deploy.

sls deploy:
```
Serverless: Load command webpack
Serverless: Invoke deploy
Serverless: Invoke package
Serverless: Building Azure Events Hooks  // note that this follows :package:
Serverless: Parsing Azure Functions Bindings.json...
Serverless: Building binding for function: hello event: httpTrigger
Serverless: Building binding for function: hello event: http
Serverless: Building binding for function: sortFood event: httpTrigger
Serverless: Building binding for function: sortFood event: http
Serverless: Invoke webpack:validate  // note that this also follows :package:
Serverless: Invoke webpack:compile
Serverless: Bundling with Webpack...
[...webpack output here...]
Serverless: Moving hello-function.json to .webpack directory.
Serverless: Moving sortFood-function.json to .webpack directory.
Serverless: Packaging service...
```
sls deploy function
```
Serverless: Load command webpack
Serverless: Invoke deploy:function
Serverless: Invoke webpack:validate  // note that this now follows :deploy:
Serverless: Invoke webpack:compile
Serverless: Bundling with Webpack...
[...webpack output here...]
Serverless: Invoke webpack:package
Serverless: Warning: No generated hello-function.json file was found. It will not be included in the package.
Serverless: Invoke package:function
Serverless: Building Azure Events Hooks  // note that this still follows :package:
Serverless: Parsing Azure Functions Bindings.json...
Serverless: Building binding for function: hello event: httpTrigger
Serverless: Building binding for function: hello event: http
Serverless: Packaging function: hello...
```

At first I thought the "Building Azure Events Hooks" was happening in the wrong spot, but now I think it's the webpack steps that do. Notice how during a full deploy, webpack follows the :package: step, but during a single-function deploy, webpack follows the :deploy: step and occurs _before_ the :package: step.

KosnIre:
Semi-related: if 2 different plugins hook into the same lifecycle event, what decides who goes first? Is there any way to ensure a specific order?

HyperBrain:
**Order**:
It is the same as with all SLS hooks - the order of the `plugins` object in the `serverless.yml` defines the order of the hooks. Serverless' PluginManager class iterates the plugins in order and pushes the hook functions serially.

**Deploy Function**:
The `deploy function` command is a completely different command with a completely different lifecycle than the `deploy` command, even if it looks like it were just an extension of the deploy command.
So `Invoke deploy:function` denotes that the deploy function command is executed, *not* deploy.
That's also the reason why the webpack plugin uses a completely separate hook for function deployment:
```
      'before:deploy:function:packageFunction': () => BbPromise.bind(this)
        .then(() => this.serverless.pluginManager.spawn('webpack:validate'))
        .then(() => this.serverless.pluginManager.spawn('webpack:compile'))
        .then(() => this.serverless.pluginManager.spawn('webpack:package')),
```
It is somehow misleading, that Serverless uses the same notation for lifecycle events for nested commands. Soin `before:deploy:function:packageFunction` the `deploy:function` is the command.
Only the last part of the event identifier is in fact the hookable lifecycle event.

HyperBrain:
Regarding the Serverless lifecycle overview, you also can have a look at my Gist https://gist.github.com/HyperBrain/50d38027a8f57778d5b0f135d80ea406.

I just noticed that I missed the deploy function lifecycles there and will add them.

KosnIre:
Ok, this is interesting :thinking: .

> It is the same as with all SLS hooks - the order of the plugins object in the serverless.yml defines the order of the hooks.

KosnIre:
> It is the same as with all SLS hooks - the order of the plugins object in the serverless.yml defines the order of the hooks.

KosnIre:
Ok, this is interesting :thinking: 

> It is the same as with all SLS hooks - the order of the plugins object in the serverless.yml defines the order of the hooks.

Thank you for that. That helps a lot. I knew something was odd, and I think I've discovered what. I had assumed that `serverless deploy function` was spawning `:package:function` just as `serverless deploy` spawns `:package:`, but that's not true. `serverless-azure-functions` [is spawning](https://github.com/serverless/serverless-azure-functions/blob/master/deploy/azureDeployFunction.js#L20-L21) `:package:function:` [to its own end](https://github.com/serverless/serverless-azure-functions/blob/master/package/azurePackageFunction.js#L18-L21).

I tried changing this: 

- I removed the .spawn() from `azureDeployFunction.js`
- In `azurePackageFunction.js`, I changed the hook to `before:deploy:function:packageFunction`, the same as used by `serverless-webpack`. 
- I ensured the listing was the correct order in my `serverless.yml` file 

And everything now works as expected.

Now the question is, did I break anything by doing this? `serverless-azure-functions` seems a little light on unit tests. @Trane9991 , I know it's been a while since you wrote this code, but can [you're listed in the git blame](https://github.com/serverless/serverless-azure-functions/blame/master/deploy/azureDeployFunction.js#L20). Was there a purpose to spawning `:package:function:` aside from getting the hook to build the event bindings? If not, I think this would be a safe change, as the hook described above accomplishes the same goal.

Trane9991:
Hey, @KosnIre 

I don't remember actual purpose of this `spawn()`, but your change looks good to me, especially if you tested the change and it works as expected 👍

KosnIre:
Thanks! This morning, I decided to be thorough and run tests with a [new template repo](https://serverless.com/framework/docs/providers/azure/guide/quick-start/) (without webpack), and now I see why the `.spawn('package:function')` is there. Without it, `serverless` will not [create the zip artifact](https://github.com/serverless/serverless/blob/master/lib/plugins/package/lib/packageService.js#L102-L108) during `sls deploy function`. To me, it feels like something serverless itself should do for us, but since it's working with it there, I'm going to just leave well-enough alone.

KosnIre:
Basic deploy testing:

HyperBrain:
> Without it, serverless will not create the zip artifact during sls deploy function.

Just verified that in Serverless' deploy function code. It is _NOT_ Serverless itself that does the spawn, but the AWS plugin, which adds the very same functionality.
So from an architectural perspective it is intended that the provider plugins spawn the `package:function` event lifecycle on their own, if they want to use Serverless' packaging. I'm quite sure that this approach was chosen to let provider plugins decide by themselves if to use the standard packaging or implement their own.

KosnIre:
> this approach was chosen to let provider plugins decide by themselves if to use the standard packaging or implement their own.

Issue #106, opened 2018-01-10
Support specifying resource group name


t-my:
From my experience, most organizations have governance and invoicing models based on resource group names.  However, currently functions deployed with sls have hard-coded resource group name:

https://github.com/serverless/serverless-azure-functions/blob/ff50c6cc35d2705d6a90c954ab727ae26359320f/provider/azureProvider.js#L45

This is clearly a major blocker for most deployments.

I could look into this, seems to be trivial to fix.

rupakg:
@t-my So you would suggest to expose it as a input param in the `serverless.yml` file? Did you get a chance to look at it?

t-my:
I I'm not familiar with contributing to serverless, so I could use a heads up here. 

Could it be simple as adding the parameter to `serverless.yml` and then something like:

` resourceGroupName = this.serverless.service.rg || ${functionAppName}-rg; `

Do I need to declare used parameters somewhere?

mikhailshilkov:
+1, I totally don't need a new resource group for each function app, I want to specify an existing one

wkirby:
Duplicates #44

Issue #48, opened 2017-03-20
Change remove to only remove the functions in the serverless.yml file


aidanmorgan:
In the case of using a custom Resource Group it seems dangerous to assume that we should delete the entire resource group when running a remove.

Change the remove operation to remove only the functions specified in the `serverless.yml` in the case of a custom resource group being used.

To allow a developer to remove the entire resource group also provide a `--clean` CLI option to the remove that will delete everything.

If no custom resource group is specified then delete the default-named resource group (current OOTB behaviour).

aidanmorgan:
Raised as part of discussion in PR  #47

PR #119, opened 2018-05-02
Update provider to allow configuring resource group

wkirby:
This PR addresses #44 and #106 by allowing you to configure `resourceGroup` and 
`deploymentName` in `serverless.yml`.

```
provider:
  name: azure
  resourceGroup: your-resource-group
  deploymentName: your-deployment-name
```

We have been using this fork in production for several weeks with no issues.

### DIFF
--- a/provider/azureProvider.js
+++ b/provider/azureProvider.js
@@ -42,8 +42,8 @@ class AzureProvider {
 
     return new BbPromise((resolve) => {
       functionAppName = this.serverless.service.service;
-      resourceGroupName = `${functionAppName}-rg`;
-      deploymentName = `${resourceGroupName}-deployment`;
+      resourceGroupName = this.serverless.service.provider.resourceGroup || `${functionAppName}-rg`;
+      deploymentName = this.serverless.service.provider.deploymentName || `${resourceGroupName}-deployment`;
 
       resolve();
     });


jeevan-shikaram:
This conflicts with the `serverless remove` command. The remove command with can be found in `./remove/azureRemove.js` deletes the resource group.

I assume the original creators recognized that if they allowed a user to assign their own resource group, there will be a danger of the resource group being deleted.

So when a user specifies a resource group, the whole resource group shouldn't be deleted as there might be other resources inside it.

Inside `./remove/lib/deleteResourceGroup.js`:
`module.exports = {
  deleteResourceGroup () {
    return this.provider.Login()
      .then(() => this.provider.DeleteDeployment())
      .then(() => this.provider.DeleteResourceGroup());
  }
};`

This should be changed to detect if the resource group user-specified and should not delete the resource group.

wkirby:
I don’t see how this conflicts, as the `remove` command still works, and this change doesn’t affect default behavior.

I take your point about `remove` being a dangerous command, but I don’t see how it is any more dangerous than the current implementation, as there’s nothing stopping me from adding to the resource group created by serverless now.

I also don’t see that this PR is a good fit for addressing the concerns raised in #48 as the feature request to configure a custom resource group name is separate from the issue raised there.

It seems to me that this should be identified in the documentation, that using a resource group that includes services not created or managed by serverless has additional risk.

ihachani:
I think the problem raised in #48  can be solved by specifying the function name and keeping track of it.

nohemirojas:
Any progress on this issue? It's still open, I have the same problem with the resource group

loket:
Same here. This is my only blocker for using Serverless Framework as I need to follow strict naming conventions when deploying to Azure.

I see the point of using the `remove` command, but that's not an issue for me.

heina-chang:
Hello,
We are working on deploying some stuff to azure and would love to be able to control the resource group.  When do you plan to merge this in?  many thanks

tbarlow12:
LGTM :shipit:

PR #140, opened 2019-05-21
ci: Add Jest testing, ESLint TypeScript & .travis.yml to run both

tbarlow12:


.eslintrc.json
```diff
@@ -1,33 +1,17 @@
...
-    "ecmaVersion": 2018,
-    "sourceType": "module",
-    "ecmaFeatures": {
-      "arrowFunctions": true,
-      "classes": true,
-      "impliedStrict": true,
-      "modules": true,
-      "spread": true
-    }
+    "@typescript-eslint/no-explicit-any": 0,
```
tbarlow12: Ignoring these rules for now, will be fixing the issues (unless ignoring them makes sense) in a future PR

package.json
```diff
@@ -6,7 +6,10 @@
   "main": "./lib/index.js",
   "author": "Azure Functions",
   "scripts": {
-    "test": "eslint src/**/*.js",
+    "lint": "eslint src/**/*.ts",
+    "lintfix": "eslint src/**/*.ts --fix",
+    "pretest": "npm run lint",
+    "test": "jest --config jest.config.json",
     "prebuild": "rm lib/ -rf && npm test",
```
mydiemho: need to change to `rm -rf lib/`, else won't work on mac (probably linux too)

mydiemho:
can you also set this to run tests on all 3 os (mac, windows, linux)?

tbarlow12:
> can you also set this to run tests on all 3 os (mac, windows, linux)?

@mydiemho I think for now, I'll just follow the convention of the other `serverless` plugins. If we run each node version on each os, that's 9 jobs, which might be annoying for the rest of the repos that are queueing up their own jobs as well

mydiemho:
> > can you also set this to run tests on all 3 os (mac, windows, linux)?
> 
> @mydiemho I think for now, I'll just follow the convention of the other `serverless` plugins. If we run each node version on each os, that's 9 jobs, which might be annoying for the rest of the repos that are queueing up their own jobs as well

@tbarlow12 I agree that that's a lot of jobs, but that is to prevent breaking changes that will ended up taking more time to fix if we don't run all the tests.  I don't think there will be too many PRs that this will become an issue.

PR #141, opened 2019-05-22
fix: Resolves issues running plugin 

wbreza:
- Fixes issue allowing plugin to function overall
- Refactored and removed dead `lib` code
- Normalized on new az login plugin and open apis
- Added typescript types to plugins


src/plugins/login/loginPlugin.ts
```diff
@@ -11,11 +11,16 @@ export class AzureLoginPlugin {
     this.provider = (this.serverless.getProvider('azure') as any) as AzureProvider;
 
     this.hooks = {
-      'before:deploy:initialize': this.login.bind(this)
+      'before:package:initialize': this.login.bind(this)
```
tbarlow12: Not entirely sure of the difference here, but the warning I got said to use `package:initialize` without the before. Guess it probably doesn't matter though, right?

src/plugins/login/loginPlugin.ts
```diff
@@ -28,7 +33,8 @@ export class AzureLoginPlugin {
     try {
       if (subscriptionId && clientId && secret && tenantId) {
         authResult = await loginWithServicePrincipalSecretWithAuthResponse(clientId, secret, tenantId);
-      } else {
+      }
```
tbarlow12: nit: Move back to same line as `}`. I know I moved this when I commented that section out

src/provider/azureProvider.ts
```diff
@@ -95,7 +84,7 @@ export default class AzureProvider {
       .on('error', () => {
         console.error('Disconnected from log streaming.');
       })
-      .on('end', this.getLogsStream.bind(this))
+      .on('end', () => this.getLogsStream(functionName))
```
tbarlow12: Advantage of arrow function over bind? Should we be using arrow function throughout?

PR #142, opened 2019-05-22
 feature: change to zipdeploy & run from package

mydiemho:
1. use the current Azure recommended zip deploy method to [run function from deployment package ](https://docs.microsoft.com/en-us/azure/azure-functions/run-functions-from-deployment-package)
1. remove support for individual function deploy since that's not supported by zipdeploy

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -13,7 +13,7 @@ export const constants = {
   scmCommandApiPath: '/api/command',
   scmDomain: '.scm.azurewebsites.net',
   scmVfsPath: '/api/vfs/site/wwwroot/',
-  scmZipApiPath: '/api/zip/site/wwwroot/'
+  scmZipDeployApiPath: '/api/zipdeploy'
 };
 
 export default constants;
--- a/src/plugins/deploy/azureDeployPlugin.ts
+++ b/src/plugins/deploy/azureDeployPlugin.ts
@@ -7,14 +7,8 @@ export class AzureDeployPlugin {
 
   constructor(private serverless: Serverless, private options: Serverless.Options) {
     this.hooks = {
-      'before:deploy:deploy': this.beforeDeploy.bind(this),
       'deploy:deploy': this.deploy.bind(this)
     };
-  }
-
-  private async beforeDeploy() {
-    // TODO: maybe this is the place to check if all relevant files exist before uploading
-    this.serverless.cli.log("NOTHING ELSE to do before deploy")
   }
 
   private async deploy() {
@@ -24,6 +18,6 @@ export class AzureDeployPlugin {
     const functionAppService = new FunctionAppService(this.serverless, this.options);
 
     const functionApp = await functionAppService.deploy();
-    await functionAppService.zipDeploy(functionApp);
+    await functionAppService.uploadFunctions(functionApp);
   }
 }
--- a/src/plugins/deploy/azureDeployPlugin.ts
+++ b/src/plugins/deploy/azureDeployPlugin.ts
@@ -16,9 +16,8 @@ export class AzureDeployPlugin {
     const functionAppService = new FunctionAppService(this.serverless, this.options);
     const functionApp = await functionAppService.get();
 
-    if (functionApp) {
-      await functionAppService.cleanUp(functionApp);
-    }
+    this.serverless.cli.log("NOTHING to do before deploy")
+    // TODO: maybe this is the place to check if all relevant files exist before uploading
   }
 
   private async deploy() {
@@ -26,9 +25,9 @@ export class AzureDeployPlugin {
     await resourceService.deployResourceGroup();
 
     const functionAppService = new FunctionAppService(this.serverless, this.options);
-    const functionApp = await functionAppService.deploy();
 
-    await functionAppService.uploadFunctions(functionApp);
-    await functionAppService.syncTriggers(functionApp);
+    // create all necessary resources: resource-group, storage account, app service plan, and app service
+    const functionApp = await functionAppService.deploy();
+    await functionAppService.zipDeploy(functionApp);
   }
 }
--- a/src/plugins/deploy/azureDeployPlugin.ts
+++ b/src/plugins/deploy/azureDeployPlugin.ts
@@ -13,11 +13,8 @@ export class AzureDeployPlugin {
   }
 
   private async beforeDeploy() {
-    const functionAppService = new FunctionAppService(this.serverless, this.options);
-    const functionApp = await functionAppService.get();
-
-    this.serverless.cli.log("NOTHING to do before deploy")
     // TODO: maybe this is the place to check if all relevant files exist before uploading
+    this.serverless.cli.log("NOTHING ELSE to do before deploy")
   }
 
   private async deploy() {
@@ -26,7 +23,6 @@ export class AzureDeployPlugin {
 
     const functionAppService = new FunctionAppService(this.serverless, this.options);
 
-    // create all necessary resources: resource-group, storage account, app service plan, and app service
     const functionApp = await functionAppService.deploy();
     await functionAppService.zipDeploy(functionApp);
   }
--- a/src/provider/armTemplates/azuredeploy.json
+++ b/src/provider/armTemplates/azuredeploy.json
@@ -103,6 +103,10 @@
                         {
                             "name": "WEBSITE_NODE_DEFAULT_VERSION",
                             "value": "8.11.1"
+                        },
+                        {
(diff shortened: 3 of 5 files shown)


src/config.ts
```diff
@@ -13,7 +13,8 @@ export const constants = {
   scmCommandApiPath: '/api/command',
   scmDomain: '.scm.azurewebsites.net',
   scmVfsPath: '/api/vfs/site/wwwroot/',
-  scmZipApiPath: '/api/zip/site/wwwroot/'
+  scmZipApiPath: '/api/zip/site/wwwroot/',
```
pjlittle: Do we still need to have the old endpoint const?

src/services/functionAppService.ts
```diff
@@ -48,68 +49,34 @@ export class FunctionAppService extends BaseService {
...
-        method: 'PUT',
-        uri: `https://${scmDomain}/api/zip/site/wwwroot/${functionName}/`,
+        method: 'POST',
+        uri: uploadUrl,
         json: true,
         headers: {
           Authorization: `Bearer ${this.credentials.tokenCache._entries[0].accessToken}`,
-          Accept: '*/*'
+          Accept: '*/*',
+          "Content-type": "application/octet-stream"
```
pjlittle: not sure if it matters to Azure, but typically multi-word headers are camel case: `Content-Type`

nitpic: Different than TypeScript, but it looks like the convention is to use single quotes vs. double quotes.

src/services/functionAppService.ts
```diff
@@ -48,68 +49,34 @@ export class FunctionAppService extends BaseService {
...
         }
       };
 
       await this._sendFile(requestOptions, functionZipFile);
```
pjlittle: Feels like we need some error handling/logging for when this fails.  I'm also curious on timeout... if you have a big enough file, will it kill the upload? Or maybe all of this is covered in the `_sendFile` function?

pjlittle:
Looking good. Does it make sense to add corresponding tests or are we going to revisit that for the deeper refactor efforts?

src/services/functionAppService.ts
mydiemho: I was wondering too what’s the best way to handle this.  _sendFile is an async function already but the zipdeploy api also let you set an isAsyncFlag (https://github.com/projectkudu/kudu/wiki/Deploying-from-a-zip-file-or-url#asynchronous-zip-deployment).  Though I’m not sure how an async within an async would work.

mydiemho:
@pjlittle we should discuss how we want to approach unit tests at standup

src/config.ts
mydiemho: Good catch, we can remove it

src/services/functionAppService.ts
tbarlow12: I think even a simple try/catch around the `await` that logs the error could be a start

src/plugins/deploy/azureDeployPlugin.ts
```diff
@@ -16,19 +16,18 @@ export class AzureDeployPlugin {
...
   private async deploy() {
     const resourceService = new ResourceService(this.serverless, this.options);
     await resourceService.deployResourceGroup();
 
     const functionAppService = new FunctionAppService(this.serverless, this.options);
-    const functionApp = await functionAppService.deploy();
 
-    await functionAppService.uploadFunctions(functionApp);
-    await functionAppService.syncTriggers(functionApp);
+    // create all necessary resources: resource-group, storage account, app service plan, and app service
```
tbarlow12: Is this comment saying that line 30 does all that? Or that this is where we _would_ do that?

mydiemho: this is saying line 30 does all that

src/plugins/deploy/azureDeployPlugin.ts
```diff
@@ -13,22 +13,17 @@ export class AzureDeployPlugin {
   }
 
   private async beforeDeploy() {
```
wbreza: Let's remove the whole hook till we have something relevant to put here.

src/services/functionAppService.ts
```diff
@@ -43,63 +44,45 @@ export class FunctionAppService extends BaseService {
...
-  public async uploadFunctions(functionApp): Promise<any> {
-    this.serverless.cli.log('Creating azure functions');
-
-    const scmDomain = functionApp.enabledHostNames[0];
+    // Upload function artifact if it exists, otherwise the full service is handled in 'uploadFunctions' method
     const functionZipFile = this.serverless.service['artifact'];
+    if (functionZipFile) {
+      this.serverless.cli.log(`-> Uploading ${functionZipFile}`);
+
+      const uploadUrl = `https://${functionAppName}${constants.scmDomain}${constants.scmZipDeployApiPath}`;
```
wbreza: I don't think these constants should be used.  In theory this should be the only function that should use this URL.  Also - the `scmDomain` is not a constant and will change based on the configuration of the function app.

src/services/functionAppService.ts
```diff
@@ -43,63 +44,45 @@ export class FunctionAppService extends BaseService {
     return response.data.value;
   }
 
-  public async deleteFunction(functionName) {
```
wbreza: We don't know if we will need `syncTriggers` or `deleteFucntion` in some other workflows.  Let's keep these methods for now.

src/services/functionAppService.ts
```diff
@@ -43,63 +44,45 @@ export class FunctionAppService extends BaseService {
...
-    this.serverless.cli.log('Syncing function triggers');
-
-    const syncTriggersUrl = `${this.baseUrl}${functionApp.id}/syncfunctiontriggers?api-version=2016-08-01`;
-    await this.sendApiRequest('POST', syncTriggersUrl);
-  }
+  async zipDeploy(functionApp) {
+    const functionAppName = functionApp.name;
+    this.serverless.cli.log(`Deploying zip file to function app: ${functionAppName}`);
 
-  public async cleanUp(functionApp) {
```
wbreza: Same with `cleanUp` and `listFunctions`.  We may need these in other workflows.

src/services/functionAppService.ts
```diff
@@ -43,63 +44,45 @@ export class FunctionAppService extends BaseService {
...
-  }
-
-  public async listFunctions(functionApp) {
-    const getTokenUrl = `${this.baseUrl}${functionApp.id}/functions?api-version=2016-08-01`;
-    const response = await this.sendApiRequest('GET', getTokenUrl);
-
-    return response.data.value || [];
-  }
-
-  public async uploadFunctions(functionApp): Promise<any> {
```
wbreza: I know we are using `zipDeploy` method for the service deployment - but I think we should keep a more generic name as the method name.  Later on when we introduce `runFromPackage` we will have multiple branches within this method.  Let's stick with `uploadFunctions` for now.

src/services/functionAppService.ts
```diff
@@ -43,63 +44,45 @@ export class FunctionAppService extends BaseService {
...
-    this.serverless.cli.log('Creating azure functions');
-
-    const scmDomain = functionApp.enabledHostNames[0];
+    // Upload function artifact if it exists, otherwise the full service is handled in 'uploadFunctions' method
     const functionZipFile = this.serverless.service['artifact'];
+    if (functionZipFile) {
```
wbreza: Invert if statement to reduce nesting.

src/services/functionAppService.ts
mydiemho: `scmDomain` is coming from `config.ts`, how will that change with the function app?

src/services/functionAppService.ts
mydiemho: just to be clear, I am adding `runFromPackage` in this PR already, it's the change in the arm template

src/provider/armTemplates/azuredeploy.json
```diff
@@ -103,6 +103,10 @@
                         {
                             "name": "WEBSITE_NODE_DEFAULT_VERSION",
                             "value": "8.11.1"
```
pjlittle: Just occurred to me that we should probably target the latest Node 8.x LTX, v8.16.0, which comes with NPM v.4.1: https://nodejs.org/en/download/releases/

I would assume this is supported?

pjlittle:
LGTM :shipit:

src/provider/armTemplates/azuredeploy.json
mydiemho: going to put this in a different pr and test :D

PR #144, opened 2019-05-23
feat: Mock of Serverless object for unit tests

tbarlow12:


src/test/mockFactory.ts
```diff
@@ -0,0 +1,80 @@
+import Serverless from 'serverless';
+import Service from 'serverless/classes/Service';
+import Utils = require('serverless/classes/Utils');
+import PluginManager = require('serverless/classes/PluginManager');
+
+export class MockFactory {
+  public static createTestServerless(service?: Service, utils?: Utils, pluginManager?: PluginManager): Serverless {
```
wbreza: Do you see a use case where we would want to manually create `utils` or `pluginManager`?  I would expect if we wanted to mock any of those calls we would just spy/mock on them after instantiation.

src/test/mockFactory.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+  }
+
+  public static createTestCli(){
+    return {
+      log: jest.fn()
+    }
+  }
+
+  public static createTestPluginManager(): PluginManager {
```
wbreza: I would change these to `private` if we don't expect to use them directly.

src/test/mockFactory.ts
```diff
@@ -0,0 +1,80 @@
+import Serverless from 'serverless';
+import Service from 'serverless/classes/Service';
+import Utils = require('serverless/classes/Utils');
+import PluginManager = require('serverless/classes/PluginManager');
+
+export class MockFactory {
+  public static createTestServerless(service?: Service, utils?: Utils, pluginManager?: PluginManager): Serverless {
+    const sls = new Serverless();
```
wbreza: We should expose the serverless `config` param in this mock factory that gets passed into `Serverless` ctor.

PR #147, opened 2019-05-23
feat: Added scaffolding for changelog

tbarlow12:
- Blatantly stolen from VoTT
- Resolves [[AB#18693]](https://dev.azure.com/dwrdev/7-11%20HA%20Serverless/_sprints/taskboard/7-11%20HA%20Serverless%20Team/7-11%20HA%20Serverless/Sprint%201)


mydiemho:
the script from VoTT was a little funky requiring `v`-prefix for source and not dest (or vice versa I can't remember).  We should change it here to be more consistent.

pjlittle:
Yeah, this is probably a good opportunity to revisit the workflow and look at using `npm version` to drive a little more consistently. I think the script is a fine place to start. We haven't talked much about branches/tags/releases, in the context of this project. What we used on VoTT feels like it will work well here.

pjlittle:
Let's chat collectively to iron out expected release workflows.

tbarlow12:
@pjlittle sounds good. We can hold off on this for now. We also should talk about how often we merge dev into master

PR #151, opened 2019-05-28
feat: Func plugin to add/remove functions within function app

tbarlow12:
- Adds CLI commands `sls func add -n {functionName}` and `sls func remove -n {functionName}`
- Adds 1 plugin with 3 hooks: (`func`, `func:add`, `func:remove`)
- Generates code directory and updates `serverless.yml` appropriately
- For now, only supports http functions. Working on supporting additional triggers
- Still working on tests, documentation and edge cases


src/index.ts
```diff
@@ -4,16 +4,20 @@ This way only one plugin needs to be added to the service in order to get access
...
-import AzureProvider from './provider/azureProvider';
-import { AzureInvoke } from './plugins/invoke/azureInvoke';
-import { AzureLogs } from './plugins/logs/azureLogs';
-import { AzureRemove } from './plugins/remove/azureRemove';
-import { AzurePackage } from './plugins/package/azurePackage';
-import { AzureDeployPlugin } from './plugins/deploy/azureDeployPlugin';
-import { AzureLoginPlugin } from './plugins/login/loginPlugin';
-import { AzureApimServicePlugin } from './plugins/apim/apimServicePlugin';
-import { AzureApimFunctionPlugin } from './plugins/apim/apimFunctionPlugin';
+import Serverless from "serverless";
```
tbarlow12: Updated to use double quotes

src/plugins/func/funcUtils.ts
```diff
@@ -0,0 +1,99 @@
+import yaml from "js-yaml";
+import fs from "fs";
+
+const functionsRegex = /functions:([\s\S]*?)\n\n/g
```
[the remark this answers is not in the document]
wbreza: Does this go back to preserving comments?  From my investigation this library doesn't support comments and don't plan on it.  Do we "really" care about preserving comments?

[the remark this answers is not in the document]
tbarlow12: The easiest way would be to just parse the yaml, update the functions section and write it back. I'm fine with that if you're ok with it

src/plugins/func/azureFunc.ts
```diff
@@ -0,0 +1,26 @@
+import Serverless from "serverless";
+
+export class AzureFuncPlugin {
```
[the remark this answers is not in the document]
tbarlow12: This was part of my investigation into sub-commands. But you're right, I think I can unify these at this point

src/plugins/func/remove/azureFuncRemove.ts
```diff
@@ -0,0 +1,57 @@
+import fs from "fs";
+import rimraf from "rimraf";
+import Serverless from "serverless";
+import { FuncPluginUtils } from "../funcUtils";
+
+export class AzureFuncRemovePlugin {
```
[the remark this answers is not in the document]
tbarlow12: Same as below

src/plugins/func/funcUtils.ts
```diff
@@ -0,0 +1,99 @@
...
+        };
+    }
+};`;
+  }
+
+  public static getFunctionJson(name: string, options: any) {
+    // TODO: This is where we would just generate function JSON from SLS object
+    // using getFunctionSlsObject(name, options). Currently defaulting to http in and out
+    return JSON.stringify({
+      "bindings": [
```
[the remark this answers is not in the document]
tbarlow12: Fine with me. Maybe have a folder of templates as we start to support other bindings. I'm envisioning a `func update` command as well

src/plugins/func/remove/azureFuncRemove.ts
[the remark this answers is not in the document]
tbarlow12: Combined in update

src/plugins/func/azureFunc.ts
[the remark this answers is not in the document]
tbarlow12: Combined in update

src/plugins/func/funcUtils.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  public static updateFunctionsYml(functionYml: any, serverlessYml?: string) {
+    serverlessYml = serverlessYml || FuncPluginUtils.getServerlessYml();
+    const newFunctionsYaml = yaml.dump(functionYml);
+    const newServerlessYaml = serverlessYml.replace(functionsRegex, `${newFunctionsYaml}\n`);
+    fs.writeFileSync("serverless.yml", newServerlessYaml);
+  }
+
+  public static getFunctionHandler(name: string) {
+    return `"use strict";
```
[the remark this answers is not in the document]
tbarlow12: So I've spent some time trying to make this work. My problem has been what kind of file to put the template in. Here are the options I've tried:

Template file options:
1. TypeScript file
   - Gets transpiled to JavaScript in build - doesn't keep template structure
2. JavaScript file
   - Can't be included in build
   - Tried setting 'allowJs' to be true, but that conflicts with 'declarations' being true
3. Text file
   - Could make it work, but it would mean an extra step post-build of copying the text file over. Not sure that's warranted just to avoid the inline script

Other options?

src/plugins/func/azureFunc.ts
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  private async func() {
+    this.serverless.cli.log("Use the func plugin to add or remove functions within Function App");
+  }
+
+  private async add() {
+    if (!("name" in this.options)) {
+      this.serverless.cli.log("Need to provide a name of function to add");
+      return;
```
mydiemho: if you return instead of throwing error, would this exit other lifecycle?

src/plugins/func/azureFunc.ts
```diff
@@ -0,0 +1,109 @@
...
+      this.serverless.cli.log(`Function ${funcToAdd} already exists`);
+      return;
+    }
+    this.createFunctionDir(funcToAdd);
+    this.addToServerlessYml(funcToAdd);
+  }
+
+  private createFunctionDir(name: string) {
+    this.serverless.cli.log("Creating function dir");
+    fs.mkdirSync(name);
```
mydiemho: should we worry about error here? in cases of permission issues?

src/plugins/func/funcUtils.ts
```diff
@@ -0,0 +1,77 @@
...
+    return JSON.stringify(httpBinding);
+  }
+
+  public static getFunctionSlsObject(name: string, options: any) {
+    return FuncPluginUtils.defaultFunctionSlsObject(name);
+  }
+
+  private static defaultFunctionSlsObject(name: string) {
+    return {
+      handler: `${name}/index.handler`,
```
mydiemho: we don't want `${name}/index.handler`, this will generate incorrect binding and function won't work.  

just `index.handler`

src/test/mockFactory.ts
```diff
@@ -37,6 +39,50 @@ export class MockFactory {
...
+    for (let i = 0; i < functionCount; i++) {
+      const functionName = `function${i+1}`;
+      data[functionName] = MockFactory.createTestFunctionMetadata(functionName)
+    }
+    return (wrap) ? {"functions": data } : data;
+  }
+
+  public static createTestFunctionMetadata(name: string) {
+    return {
+      "handler": `${name}/index.handler`,
```
mydiemho: same as above no `${name}`

src/plugins/func/funcUtils.ts
tbarlow12: This is the structure that `serverless.yml` needs to find the handler since it will live inside the directory `{functionName` and inside the file `index.js`, with the function name `handler`

src/test/mockFactory.ts
tbarlow12: Same as above

src/plugins/func/azureFunc.ts
tbarlow12: I'll put a try catch here

src/plugins/func/azureFunc.ts
tbarlow12: What other lifecycle are you referring to? This will just exit the process, and they would need to run again with a name specified

README.md
```diff
@@ -18,6 +18,26 @@ This plugin enables Azure Functions support within the Serverless Framework.
...
 
+### Creating or removing Azure Functions
+
+To create a new Azure Function within your function app, run the following command from within your app's directory:
+
+```bash
+sls func add -n {functionName}
+```
+
+This will create a new `{functionName}` directory at the root of your application with `index.js` and `function.json` inside the directory. It will also update `serverless.yml` to contain the new function.
```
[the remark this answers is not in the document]
tbarlow12: As I've learned more about the tool, I don't know that we need a sample.dat. It's pretty basic and kind of confusing as to what it actually does. Was going to talk to the team about removing it from the template

src/plugins/func/azureFunc.ts
mydiemho: ah you're right, I was thinking of other command with more lifecyles

src/plugins/func/funcUtils.ts
mydiemho: yes but the plugin will generate incorrect `scriptFilePath` since `function.json` now resides at the same level as `index.json`

src/test/mockFactory.ts
mydiemho: same as above :p

README.md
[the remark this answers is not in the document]
mydiemho: I have no strong preference, whatever you think is best :D

src/plugins/func/funcUtils.ts
```diff
@@ -0,0 +1,77 @@
...
+    return JSON.stringify(httpBinding, null, 2);
+  }
+
+  public static getFunctionSlsObject(name: string, options: any) {
+    return FuncPluginUtils.defaultFunctionSlsObject(name);
+  }
+
+  private static defaultFunctionSlsObject(name: string) {
+    return {
+      handler: `index.handler`,
```
mydiemho: Nit: can change back to double quote since we're no longer interpolating a variable

src/shared/utils.ts
```diff
@@ -110,4 +111,11 @@ export class Utils {
   
     return metaData;
   }
+
+  public static interpolateFile(sls: Serverless, path: string, params: Map<string, string>) {
+    const template = sls.utils.readFileSync(path);
+    const names = params.keys();
+    const vals = params.values();
+    return new Function(...names, `return \`${template}\`;`)(...vals);
```
mydiemho: Q: can 
```
`return \`${template}\`;`
```
be replace with 
```
`return ${template};`
```

tbarlow12: This is actually a string interpolation that won't happen here. We actually have the same function (kind of) in VoTT. This is expecting a file with templated variables. I'm actually not using this right now... I was using it when I had a javascript template file. But I might keep it in just in case we go back to that

src/plugins/func/funcUtils.ts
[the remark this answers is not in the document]
PIC123: Yah, the ideal solution would be to be able to include a js file of the handler since that's technically what we're packaging, but if that isn't really an option, then inline seems reasonable.

src/plugins/func/azureFunc.test.ts
```diff
@@ -0,0 +1,117 @@
...
+      }, {createCwd: true, createTmp: true});    
+    });
+  
+    afterAll(() => {
+      mockFs.restore();
+    });
+  
+    it("returns with missing name", async () => {
+      const sls = MockFactory.createTestServerless();
+      const options = MockFactory.createTestServerlessOptions();
```
PIC123: Q: Would it make sense to add a default name to the options and then remove/override as needed? Just thinking if we'll need it for other function tests if we add more functionality, like the option to include the extension bundles

src/plugins/func/funcUtils.ts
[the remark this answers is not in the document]
wbreza: Let's add a work item to re-review this later.

src/plugins/func/azureFunc.test.ts
tbarlow12: For sure. Let's add that in another PR

src/plugins/func/funcUtils.ts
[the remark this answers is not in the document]
tbarlow12: [AB#18775]

PR #152, opened 2019-05-29
test: Unit tests for ResourceService and LoginService

tbarlow12:
Resolves [AB#18738] and [AB#18740]


src/test/mockFactory.ts
```diff
@@ -7,11 +7,11 @@ import PluginManager = require("serverless/classes/PluginManager");
 export class MockFactory {
   public static createTestServerless(config?: any): Serverless {
     const sls = new Serverless(config);
-    sls.service = MockFactory.createTestService();
-    sls.utils = MockFactory.createTestUtils();
-    sls.cli = MockFactory.createTestCli();
-    sls.pluginManager = MockFactory.createTestPluginManager();
-    sls.variables = {};
+    sls.service = config && config.service || MockFactory.createTestService();
```
wbreza: I think i know what you are trying to do here, but I don't believe it is working as you expect.  For example - if `config.service` is the defined then `sls.service` will be set to `true` not the value of `config.service`, otherwise it will use the `MockFactory.createTestService()`.

src/services/loginService.test.ts
```diff
@@ -0,0 +1,42 @@
...
+  beforeAll(() => {
+    // Because the functions we use for authentication are exported
+    // as functions from their respective modules 
+    // (open, interactiveLoginWithAuthResponse and 
+    // loginWithServicePrincipalSecretWithAuthResponse),
+    // it is extremely difficult to mock their functionality.
+    // As a workaround, they have been placed in the thinnest possible
+    // functions within the Azure Login service, which we will
+    // use to make assertions on the functionality of the login service itself
+    AzureLoginService.interactiveLogin = jest.fn();
```
wbreza: Another idea would be to mock the whole incoming module using `jest.mock`.  Virtually the same behavior so I am fine with this approach.

src/test/mockFactory.ts
tbarlow12: Makes sense. I'll just add a ternary operator instead. Thanks

PR #155, opened 2019-05-31
test: Unit tests for Azure Index

tbarlow12:
Resolves [AB#18742]


src/index.ts
```diff
@@ -15,7 +15,7 @@ import { AzureLoginPlugin } from "./plugins/login/loginPlugin";
 import { AzureApimServicePlugin } from "./plugins/apim/apimServicePlugin";
 import { AzureApimFunctionPlugin } from "./plugins/apim/apimFunctionPlugin";
 
-export class AzureIndex {
+export default class AzureIndex {
```
wbreza: Does adding this `default` screw up any of the exports that serverless cares about?  We initially saw some weirdness trying to use standard ES6 module export features with the root sls plugin.

tbarlow12: It seems like as long as we have the `module.exports` at the end, it still works. Ran some deploys with this, and it seemed to work ok

PR #156, opened 2019-05-31
test: Unit tests for Function App Service

tbarlow12:
Resolves [AB#18737]


src/plugins/login/loginPlugin.test.ts
```diff
@@ -56,6 +57,7 @@ describe("Login Plugin", () => {
     AzureLoginService.servicePrincipalLogin = servicePrincipalLogin;
     
     const sls = MockFactory.createTestServerless();
+    delete sls.variables['azureCredentials']
```
wbreza: Can we move these into a `beforeEach` or similar to keep it DRY?

src/services/baseService.ts
```diff
@@ -24,6 +24,10 @@ export abstract class BaseService {
     }
   }
 
+  protected log(message: string) {
```
wbreza: Nice :)

src/test/mockFactory.ts
```diff
@@ -37,9 +37,17 @@ export class MockFactory {
     }
   }
 
-  private static createTestService(): Service {
+  public static createTestFunctionApp() {
     return {
```
wbreza: We should add return typings here.

src/services/functionAppService.test.ts
```diff
@@ -0,0 +1,165 @@
...
+    mockFs({
+      "app.zip": "contents",
+    }, {createCwd: true, createTmp: true});
+  });
+
+  beforeEach(() => {
+    webAppDelete = jest.fn();
+    sendFile = jest.fn();
+
+    WebSiteManagementClient.prototype.webApps = {
```
wbreza: These methods can be mocked by mocking the sub object directly.  ex) `WebApp.prototype.get = jest.fn(...)`

src/services/functionAppService.test.ts
```diff
@@ -0,0 +1,165 @@
...
+  const masterKeyUrl = `https://${app.defaultHostName}/admin/host/systemkeys/_master`;
+  const authKeyUrl = `https://management.azure.com${app.id}/functions/admin/token?api-version=2016-08-01`;
+  const syncTriggersUrl = `https://management.azure.com${app.id}/syncfunctiontriggers?api-version=2016-08-01`;
+  const listFunctionsUrl = `https://management.azure.com${app.id}/functions?api-version=2016-08-01`;
+  const uploadUrl = `https://${app.name}${constants.scmDomain}${constants.scmZipDeployApiPath}`
+
+  beforeAll(() => {   
+
+    // TODO: How to spy on defaul exported function?
+    const axiosMock = new MockAdapter(axios);
```
wbreza: Do you find this mock adapter useful?  I have mixed feelings.

src/test/mockFactory.ts
```diff
@@ -51,8 +59,42 @@ export class MockFactory {
...
       custom: null,
-      provider: {} as any,
-    };
+      provider: MockFactory.createTestAzureServiceProvider(),
+      service: "serviceName",
+      artifact: "app.zip",
+    } as any as Service;
+  }
+
+  public static createTestFunctions(functionCount = 3) {
```
wbreza: We should add typings to these mock factory functions.

src/services/functionAppService.test.ts
tbarlow12: Yes and no... If I could make assertions on it, it would be amazing. I had to do something fairly hacky to validate the `syncTriggers`. I wasn't able to spy on axios directly. Maybe we can chat more about how to do that

src/test/mockFactory.ts
tbarlow12: Part of the reason for not having typings is my own not knowing what types they are 😆 

But agreed. I'll do some more digging here

src/plugins/login/loginPlugin.test.ts
tbarlow12: The only issue with that is the deletion of the credentials is specific to these tests only. Other tests depend on the credentials still being there

src/models/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from "./azureProvider";
```
wbreza: Why the exports?

src/models/serverless.ts
```diff
@@ -0,0 +1,8 @@
+export interface ServerlessYml {
```
wbreza: I would remove `Yml` and maybe rename to `ServerlessAzureConfig` or similar.

src/plugins/package/azurePackage.test.ts
```diff
@@ -11,9 +11,6 @@ describe("Azure Package Plugin", () => {
   it("sets up provider configuration", async () => {
     const slsFunctionConfig = MockFactory.createTestSlsFunctionConfig();
     const sls = MockFactory.createTestServerless();
-    Object.assign(sls.service, {
```
wbreza: How do the functions config set if you are removing this?

src/services/functionAppService.ts
```diff
@@ -30,7 +30,7 @@ export class FunctionAppService extends BaseService {
     return response;
   }
 
-  public async getMasterKey(functionApp) {
+  public async getMasterKey(functionApp?) {
```
wbreza: Add typeref for `functionApp`

src/plugins/package/azurePackage.test.ts
wbreza: Looks like with the updates to mock factory you need to now pass into `createTestServerless` ??

tbarlow12: For this instance in particular, since it's just using the `createTestSlsFunctionConfig` out of the box, that will already be in the `sls` object. If there is a different function config to pass, that can be achieved by doing:
```
createTestServerless({
   service: createTestService(functionConfig)
})
```

src/models/index.ts
tbarlow12: To make model imports easy. I can remove these

src/services/apimService.test.ts
```diff
@@ -37,18 +38,19 @@ describe("APIM Service", () => {
 
   beforeEach(() => {
     const slsConfig: any = {
+      ...MockFactory.createTestService(MockFactory.createTestSlsFunctionConfig()),
```
tbarlow12: @wbreza this is another option for using a custom config

PR #157, opened 2019-06-01
feat: Azure API management

wbreza:
- Adds better logging for API management (displays function urls)
- Adds unit tests
- Updates configuration for better Jest debugging


src/test/mockFactory.ts
```diff
@@ -22,8 +23,23 @@ export class MockFactory {
       noDeploy: null,
       region: null,
       stage: null,
-      watch: null
-    }
+      watch: null,
+    };
+  }
+
+  public static createTestArmSdkResponse<R>(model: any, statusCode: number): Promise<R> {
```
tbarlow12: Nice. Q: Deciding between `<T>` and `<R>`? What's the convention?

src/test/responses/apim-get-api-404.json
```diff
@@ -0,0 +1,7 @@
+{
```
tbarlow12: Nice

package.json
```diff
@@ -4,13 +4,16 @@
   "description": "Provider plugin for the Serverless Framework v1.x which adds support for Azure Functions.",
   "license": "MIT",
   "main": "./lib/index.js",
+  "typings": "./lib/index.d.ts",
   "author": "Azure Functions",
   "scripts": {
     "lint": "eslint src/**/*.ts",
```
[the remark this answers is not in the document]
wbreza: Still figuring that out.  I was seeing that VS code was giving me a couple errors with TS lint installed.  Our ES lint config was not being respected, so I added a ts lint config to mitigate.

src/test/mockFactory.ts
wbreza: `<T>` is typically used to reference some incoming types where as `<R>` is used for outgoing / return types.

src/test/responses/apim-get-api-404.json
wbreza: Yes - this is one idea for some of the HTTP mocks where we could have real templated responses vs having to set it all up in code.

src/services/apimService.ts
```diff
@@ -2,62 +2,101 @@ import Serverless from 'serverless';
...
  */
 export class ApimService extends BaseService {
   private apimClient: ApiManagementClient;
   private functionAppService: FunctionAppService;
-  private config: any;
-  
-  constructor(serverless: Serverless, options: Serverless.Options) {
+  private config: ApiManagementConfig;
+
+  constructor(serverless: Serverless, options?: Serverless.Options) {
```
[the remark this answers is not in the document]
wbreza: It's optional and right now there aren't any options that are used in the service.

package.json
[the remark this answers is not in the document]
wbreza: Our tslint & eslint configs should now be all set.  This PR also includes a `.vscode/settings` config to allow the eslint plugin to work as expected.

src/services/apimService.ts
```diff
@@ -67,73 +106,89 @@ export class ApimService extends BaseService {
...
+        this.resourceGroup,
+        this.config.name,
+        this.config.api.name,
+        options.function,
+        operationConfig,
+      );
+
+      await client.apiOperationPolicy.createOrUpdate(this.resourceGroup, this.config.name, this.config.api.name, options.function, {
         format: 'rawxml',
         value: `
```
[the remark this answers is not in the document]
wbreza: Not sure how much this would need to change.  Let's keep it for now.

package.json
```diff
@@ -41,20 +45,22 @@
     "rimraf": "^2.6.3"
   },
   "devDependencies": {
+    "@babel/runtime": "^7.4.5",
```
pjlittle: Guessing that Babel is required for jest? Something that we were previously getting 'for free' w/ `create-react-app`?

wbreza: Yes - instead of doing a straight tsc compile for jest we are using babel.  We are getting a better jest debugging experience doing this for some reason that I don't fully understand.

PR #158, opened 2019-06-04
ci: Added NPM deployment info to .travis.yml

tbarlow12:


.travis.yml
```diff
@@ -3,17 +3,26 @@ language: node_js
...
-  - npm install
-
+- npm install
 script:
-  - npm test
+- npm test
 
 after_success:
-  - cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js && rm -rf ./coverage
+- cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js && rm -rf ./coverage
```
wbreza: Does this push the coverage report up to coveralls automatically?

.travis.yml
```diff
@@ -3,17 +3,26 @@ language: node_js
...
 
 sudo: false
 
 install:
-  - npm install
-
+- npm install
 script:
-  - npm test
+- npm test
```
wbreza: should we also add linting verification? Or is that a `pre` call on `test`

tbarlow12: Exactly. Now that you mention it though, we should be calling `npm run build` so that we run `tsc` as well. Currently, `test` runs as a precursor to `build`. I think I saw that change in your incoming PR, so maybe we can update whichever one comes in last

.travis.yml
tbarlow12: Yes, it seems automagic, but our coverage badge is still `unknown`, and I'm guessing that's because we haven't had it run on `master` yet.

pjlittle: @mydiemho might know, but my guess is that it targets a default branch (master). I wonder if that's something we can change? Everything will get there eventually...

PR #159, opened 2019-06-05
feat: Azure Functions Offline plugin

tbarlow12:
- Added `AzureOfflinePlugin`
- Added `OfflineService` that contains the logic
- Various questions and topics for discussion within team, will post those on each line

Update:
- Added fixes to `func` plugin to account for new package structure
- Added `FuncService` to contain logic and utility functions for plugin


src/plugins/offline/azureOfflinePlugin.ts
```diff
@@ -0,0 +1,48 @@
...
+
+
+export class AzureOfflinePlugin {
+  public hooks: { [eventName: string]: Promise<any> };
+  public commands: any;
+  private offlineService: OfflineService;
+
+  public constructor(private serverless: Serverless, private options: Serverless.Options) {
+    this.hooks = {
+      "before:offline:offline": this.azureOfflineBuild.bind(this),
```
tbarlow12: Running build before `offline` and as it's own command `offline build`

src/services/baseService.ts
```diff
@@ -11,15 +12,15 @@ export abstract class BaseService {
   protected resourceGroup: string;
   protected deploymentName: string;
 
-  public constructor(protected serverless: Serverless, protected options: Serverless.Options) {
+  public constructor(protected serverless: Serverless, protected options: Serverless.Options, authenticate = true) {
```
tbarlow12: Added this for services like `ConverterService` or `OfflineService` that could benefit from shared service code, but that don't need to authenticate with Azure

src/services/baseService.ts
```diff
@@ -83,4 +84,12 @@ export abstract class BaseService {
         }));
     });
   }
+
+  protected getServerlessConfig(): ServerlessConfig {
+    return this.serverless.utils.readFileSync("serverless.yml") as ServerlessConfig;
```
tbarlow12: There is probably an easier way to do this with the Serverless Utils, but I haven't dug deep enough yet

wbreza: Where is this used? What part of the config is required?  Most of this is just on the `serverless` object itself, no?

tbarlow12: Need to investigate further here. With the new debugging tools, that should be a lot easier :)

src/plugins/offline/azureOfflinePlugin.ts
tbarlow12: It's different - seems to be much more verbose than what we would need. My vision here is (and please let me know if you have a different one):

```
sls offline build     (run the conversion put the proper metadata in the respective function.json files)
sls offline      (run the build and then spawn the `npm start` process)
```

src/plugins/offline/azureOfflinePlugin.ts
```diff
@@ -0,0 +1,45 @@
...
+              "build",
+            ]
+          }
+        }
+      }
+    }
+  }
+
+  private async azureOfflineBuild(){
+    this.offlineService = new OfflineService(this.serverless, this.options);
```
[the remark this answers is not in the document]
tbarlow12: Good call. Will do.

src/plugins/offline/azureOfflinePlugin.ts
tbarlow12: Maybe the offline plugin just does the build and the docs just say to run `npm start`. Open to suggestions here

src/shared/utils.ts
```diff
@@ -95,15 +95,17 @@ export class Utils {
...
+  public static getEntryPointAndHandlerPath(handler: string) {
     let handlerPath = "handler.js";
     let entryPoint = handler;
     const handlerSplit = handler.split(".");
+    const slashIndex = handler.lastIndexOf("/");
 
     if (handlerSplit.length > 1) {
       entryPoint = handlerSplit[handlerSplit.length - 1];
-      handlerPath = `${handler.substring(0, handler.lastIndexOf("."))}.js`;
+      handlerPath = `${handler.substring(slashIndex + 1, handler.lastIndexOf("."))}.js`;
```
tbarlow12: Currently, the `function.json` file looks something like this:
```
{
   ...
   scriptFile: {functionName}/index.js
   ...
}
```
This causes the file not to be found because it uses a relative path from `function.json`, which sits alongside `index.js`. This fix simply trims the directory name from the path, resulting in just `index.js`.

@wbreza I know you have an open PR for packaging updates. Not sure if you ran into this at all or if you have any other thoughts

src/services/baseService.ts
tbarlow12: Update: `ConverterService` no longer part of this. Using normal packaging process to build `function.json` files

src/services/baseService.ts
```diff
@@ -11,15 +12,15 @@ export abstract class BaseService {
...
+  public constructor(protected serverless: Serverless, protected options: Serverless.Options, authenticate = true) {
     this.baseUrl = "https://management.azure.com";
     this.serviceName = serverless.service["service"];
     this.credentials = serverless.variables["azureCredentials"];
     this.subscriptionId = serverless.variables["subscriptionId"];
     this.resourceGroup = serverless.service.provider["resourceGroup"] || `${this.serviceName}-rg`;
     this.deploymentName = serverless.service.provider["deploymentName"] || `${this.resourceGroup}-deployment`;
 
-    if (!this.credentials) {
+    if (!this.credentials && authenticate) {
```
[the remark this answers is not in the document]
tbarlow12: @wbreza one thought is to create another abstract class called `AzureService` that inherits from `BaseService`. `AzureService` would always require auth and would contain all Azure-related information

src/services/funcService.ts
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  private removeFromServerlessYml(functionName: string) {
+    const functions = this.slsFunctions();
+    delete functions[functionName];
+    this.updateFunctionsYml(functions)
+  }
+
+  private getServerlessYml() {
+    return this.serverless.utils.readFileSync("serverless.yml");
```
pjlittle: Does the framework have a way to get the name of the yaml file? With the new `--config` option, it's possible that the file won't be named `serverless.yml`. We don't have to over-index around this, but just something to keep in mind.

src/services/funcService.ts
```diff
@@ -0,0 +1,132 @@
...
+  private updateFunctionsYml(functionYml: any) {
+    const serverlessYml = this.getServerlessYml();
+    serverlessYml["functions"] = functionYml;
+    this.serverless.utils.writeFileSync("serverless.yml", yaml.dump(serverlessYml));
+  }
+
+  private getFunctionHandler(name: string) {
+    return `"use strict";
+
+module.exports.handler = async function (context, req) {
```
pjlittle: I almost want to suggest making this a separate template/file (also easier to tell if it's valid code), but given that it's encapsulated here and relatively short, I'm okay with it.

pjlittle:
Looking good - just a couple comments. Great work on the tests. :+1:

src/services/funcService.ts
tbarlow12: Good point. Wasn't accounting for this. I'll investigate further here.

tbarlow12: Added a getter that defaults to `serverless.yml`. Good callout

src/services/funcService.ts
tbarlow12: I'd love to capture this in a template file, and we actually discussed this in a previous PR (previous PR got a little murky with all the refactoring, so I closed and opened a new one). The issue I ran into was deciding what kind of file to store the template in:

```
1. .ts - transpiles to messy .js in build
2. .js - can't be included with current tsconfig (`allowJs` and `declarations` attribute cannot both be true)
3. .txt - Would need to copy to file post-build to the lib directory - seems heavy handed for this ask
```
If you have any other thoughts here, I'm open to any and all suggestions :)

pjlittle: yeah, good call. I was thinking some way of reading from a file, but it gets messy in the context of the rest of the build. Will keep thinking about it -- would be nice to have separate, but we can call it good until something else better presents itself.

src/plugins/offline/azureOfflinePlugin.test.ts
```diff
@@ -0,0 +1,69 @@
...
+  });
+
+  it("invokes offline hook", async () => {
+    const sls = MockFactory.createTestServerless();
+    const plugin = createPlugin(sls);
+    await invokeHook(plugin, "offline:offline");
+    // Trivial test for now. In the future, this process
+    // may spawn the start process itself rather than telling
+    // the user how to do it.
+    expect(sls.cli.log).toBeCalledTimes(3);
```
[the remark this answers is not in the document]
tbarlow12: Yeah, holding off on the spawn for now. Hoping `npm start` is not too much work for the time being

PR #164, opened 2019-06-11
feat: Adds webpack support for azure plugin

wbreza:
When azure serverless project is deploy with webpack we need to ensure that the required files for azure functions are part of the deployment package.

- Copies host.json & function.json files for each serverless function.


src/plugins/login/loginPlugin.test.ts
```diff
@@ -19,7 +19,7 @@ describe("Login Plugin", () => {
     const options = MockFactory.createTestServerlessOptions();
```
mydiemho: this login order change has nothing to do with webpack, I would put this in a different pr

src/plugins/login/loginPlugin.ts
```diff
@@ -10,7 +10,7 @@ export class AzureLoginPlugin {
     this.provider = (this.serverless.getProvider("azure") as any) as AzureProvider;
 
     this.hooks = {
```
mydiemho: same as above

src/plugins/package/azurePackage.ts
```diff
@@ -1,54 +1,44 @@
...
+import { PackageService } from "../../services/packageService";
 
 export class AzurePackage {
+  private bindingsCreated: boolean = false;
+  private packageService: PackageService;
   public provider: AzureProvider;
   public hooks: { [eventName: string]: Promise<any> };
 
-  public constructor(private serverless: Serverless, private options: Serverless.Options) {
+  public constructor(private serverless: Serverless) {
```
tbarlow12: Reason for no options?

wbreza: The plugin doesn't use any options right now - so removed it.  If it starts to take options in the future I'm fine bringing it back.

src/plugins/login/loginPlugin.test.ts
wbreza: Reverted login changes.

src/plugins/login/loginPlugin.ts
wbreza: Reverted login changes.

PR #166, opened 2019-06-13
feat: Deploy APIM CORS Policy

wbreza:
When the APIM CORS policy is defined within the serverless yaml then we need to deploy the XML policy document during deployment.

Resolve ABA#207


src/models/apiManagement.ts
```diff
@@ -4,9 +4,18 @@ export interface ApiManagementConfig {
   name: string;
   api: ApiContract;
   backend?: BackendContract;
+  cors?: ApiCorsPolicy;
 }
 
 export interface ApiOperationOptions {
```
mydiemho: we should start using jsdocs

src/services/apimService.ts
```diff
@@ -252,7 +252,73 @@ export class ApimService extends BaseService {
...
       this.log("Error creating APIM Property");
+      this.log(JSON.stringify(e.body, null, 4));
       throw e;
     }
   }
+
+  /**
+   * Creates the XML payload that defines the API operation policy to link to the configured backend
+   */
+  private createApiOperationXmlPolicy(): string {
```
mydiemho: should we have tests to verify this object data format?

src/services/apimService.ts
```diff
@@ -252,7 +252,73 @@ export class ApimService extends BaseService {
...
+   * Creates the XML payload that defines the specified CORS policy
+   * @param corsPolicy The CORS policy
+   */
+  private createCorsXmlPolicy(corsPolicy: ApiCorsPolicy): string {
+    const origins = corsPolicy.allowedOrigins ? corsPolicy.allowedOrigins.map((origin) => ({ origin })) : null;
+    const methods = corsPolicy.allowedMethods ? corsPolicy.allowedMethods.map((method) => ({ method })) : null;
+    const allowedHeaders = corsPolicy.allowedHeaders ? corsPolicy.allowedHeaders.map((header) => ({ header })) : null;
+    const exposeHeaders = corsPolicy.exposeHeaders ? corsPolicy.exposeHeaders.map((header) => ({ header })) : null;
+
+    const policy = [{
```
mydiemho: same with this

src/services/apimService.ts
wbreza: This would need to be an e2e integration test to the azure API, otherwise I've validated the shape of the xml document.

src/models/apiManagement.ts
wbreza: Added js docs to APIM models.

PR #170, opened 2019-06-19
docs: Design documents for rollback functionality

tbarlow12:


docs/DEPLOY.md
```diff
@@ -0,0 +1,19 @@
+# Deploy
+
+##### `sls deploy`
+- Description: Deploy resource group containing function app and other resources. Also deploys Function App code
+- Two Approaches:
+  - Deploy resource group, upload packaged artifact directly to function app. Sets function app `RUN_FROM_PACKAGE` setting to `1`.
+  - Deploy resource group, upload packaged function app to blob storage with version name. Sets function app `RUN_FROM_PACKAGE` setting to path of zipped artifact in blob storage
+- Open Questions:
+  - Default to deploy to blob storage? Default to zip deploy directly to function app?
```
mydiemho: My vote is on zip deploy default to function app and not blob storage.  This default will be more useful for the general user so they don't incur extra cost maintaining a bloc storage.

docs/DEPLOY.md
```diff
@@ -0,0 +1,19 @@
...
+  - Container name for code artifacts? Hard coded? Configurable? Default to ARTIFACTS?
```
mydiemho: `ARTIFACTS` default make sense, but also provide customization in `serverless.yml`

docs/DEPLOY.md
```diff
@@ -0,0 +1,19 @@
+# Deploy
+
+##### `sls deploy`
+- Description: Deploy resource group containing function app and other resources. Also deploys Function App code
+- Two Approaches:
+  - Deploy resource group, upload packaged artifact directly to function app. Sets function app `RUN_FROM_PACKAGE` setting to `1`.
+  - Deploy resource group, upload packaged function app to blob storage with version name. Sets function app `RUN_FROM_PACKAGE` setting to path of zipped artifact in blob storage
```
mydiemho: is this one blob storage per resource group?

docs/ROLLBACK.md
```diff
@@ -0,0 +1,16 @@
...
+##### `sls rollback`
+- Description - Roll back deployment of Function App & Resource Group
+- Options:
+  - `-t` or `--timestamp` - Timestamp associated with version to target
+  - `-v` or `--version` - Version of function app
+  - Use `sls deploy list` to discover timestamps and versions
+  - Defaults to previous deployment and version number
+- Important to note that there is no option for rolling back an individual function. A function app is considered one unit and will be rolled back as such.
+- Open questions:
+  - How to ensure rollback to corresponding ARM deployment and function app from blob storage? Custom named deployments containing version number?
```
mydiemho: does this command need to know about stage and region?

docs/DEPLOY.md
tbarlow12: Sounds good. I think that would just mean that `rollback` would not be an option. Something to be aware of is that a function app basically requires a blob storage to be deployed alongside it

docs/ROLLBACK.md
tbarlow12: Yes. Good callout. Would use the same convention as deployment

docs/DEPLOY.md
tbarlow12: With the ARM template as is, there will be one blob storage account deployed per resource group. Seems like it's required for function apps. Don't know how hard that *required* is, but I've yet to see a function app deployed without a storage account deployed alongside it. @wbreza any insight here?

tbarlow12:
image

tbarlow12:
image

docs/DEPLOY.md
pjlittle: Does serverless let you specify the name of the S3 bucket?  My hunch is no... that being said, I don't have a huge preference one way or the other.

docs/DEPLOY.md
```diff
@@ -0,0 +1,39 @@
...
+  participant f as Function App
+  participant b as Blob Storage
+
+  note right of s: `sls deploy`
+  s ->> r: Create resource group
+  s ->> r: Deploy ARM template
+  r ->> f: Included in ARM template
+  r ->> b: Included in ARM template
+  note right of s: Zip code
+  s ->> b: Deploy zip code with name {appName}-v{version}.zip
```
pjlittle: I think we're going to need more than version -- it's likely why serverless elected to use timestamp in the AWS implementation. Something like short commit SHA could be a good candidate and likely more meaningful -- you can always get a timestamp from blob storage.

pjlittle:
I like this approach and it's about the closest we'll get to the AWS experience today.  That being said, I'm still hung up on leveraging source control and CI/CD to effectively control going forwards or backwards with changes. Still, I see the value and elegance in using blob effectively as an 'artifactory' of releases.

I almost want to suggest having a param in the serverless yaml that would toggle between zip deploy and run from blob, like @mydiemho suggested and perhaps default to zip deploy.  

*Open Question:* if an Azure function is running from blob and you update/change the zip file in blob, will the new file eventually be used/referenced by the function? Maybe run in mixed mode while containers are cycled in and out? Could be another argument for having another unique value in the zip name.

wbreza:
Please update the docs based on decisions that we all discussed.

tbarlow12:
> @tbarlow12 Please update the docs based on decisions that we all discussed.

@wbreza will do. Thanks for the reminder

docs/ROLLBACK.md
```diff
@@ -0,0 +1,55 @@
...
+-----------
+Name: myFunctionApp-t1561479444
+Timestamp: 1561479444
+Datetime: 2019-06-25T16:17:24+00:00
+-----------
+
+$ sls rollback -t 1561479506
+```
+
+##### Sequence diagram for rollback
```
mydiemho: can you add a screenshot of the diagram from the comment?

docs/DEPLOY.md
```diff
@@ -56,3 +56,64 @@ then user try to deploy
...
+  rollback: true
+  # Container in blob storage containing deployed packages
+  # Default is DEPLOYMENT_ARTIFACTS
+  container: MY_CONTAINER_NAME
+
+functions:
+  ...
+```
+
+If rollback is enabled, the name of the created package will include the UTC timestamp of its creation. This timestamp will also be included in the name of the Azure deployment so as to be able to link the two together. Both names will be with `{serviceName}-t{timestamp}` (with `.zip` in the case of the packaged code).
```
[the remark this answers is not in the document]
tbarlow12: Yes, this might be redundant. I will clarify here

docs/ROLLBACK.md
tbarlow12: It will get generated as part of the markdown

PR #172, opened 2019-06-19
feat: Azure region & stage defaults

wbreza:
Ensures valid Azure values are set for region and stage.  Renames instances of "location" to "region" to align with core naming convention.


src/plugins/deploy/azureDeployPlugin.ts
```diff
@@ -12,6 +12,11 @@ export class AzureDeployPlugin {
   }
 
   private async deploy() {
+    this.serverless.cli.log("OPTIONS");
```
wbreza: Will Remove

tbarlow12: I'll assume you mean lines 15 & 16 here

src/services/baseService.test.ts
```diff
@@ -21,6 +21,14 @@ class TestService extends BaseService {
     return this.sendFile(requestOptions, filePath);
   }
 
+  public getSlsRegion() {
```
tbarlow12: Seems like pretty thin wrappers for these two functions. Any reason for having both `getSlsRegion` and `getRegion`?

src/services/functionAppService.test.ts
```diff
@@ -29,7 +28,7 @@ describe("Function App Service", () => {
   const authKeyUrl = `${baseUrl}${app.id}/functions/admin/token?api-version=2016-08-01`;
   const syncTriggersUrl = `${baseUrl}${app.id}/syncfunctiontriggers?api-version=2016-08-01`;
   const listFunctionsUrl = `${baseUrl}${app.id}/functions?api-version=2016-08-01`;
-  const uploadUrl = `https://${app.enabledHostNames[0]}${constants.scmZipDeployApiPath}/`
+  const uploadUrl = `https://${app.enabledHostNames[1]}/api/zipdeploy/`;
```
tbarlow12: Why the switch from `enabledHostNames[0]` to `enabledHostNames[1]`? Might be good to document, even if it's just in the test

src/shared/utils.ts
```diff
@@ -113,6 +113,10 @@ export class Utils {
     return metaData;
   }
 
+  public static prettyPrint(json) {
```
tbarlow12: Pretty :)

src/test/mockFactory.ts
```diff
@@ -184,7 +184,7 @@ export class MockFactory {
     const data = {
       provider: {
         name: "azure",
-        location: "West US 2"
+        region: "West US 2"
```
tbarlow12: Good switch. Location was confusing

src/plugins/deploy/azureDeployPlugin.ts
wbreza: I was referring to all the log statements.  I put those in for debugging.

src/services/baseService.test.ts
wbreza: These are purely in a `TestService` class purposely built to test the core/base functionality of the `BaseService`

src/services/functionAppService.test.ts
wbreza: I hardened the retrieval of the SCM domain from the list of domains in the website.  Based on that update I also changed the mock data used in the test and know that the SCM domain is the 2nd one in this case.

wbreza: Updated test to do similar logic.

src/services/resourceService.ts
```diff
@@ -14,11 +14,9 @@ export class ResourceService extends BaseService {
   public async deployResourceGroup() {
     this.log(`Creating resource group: ${this.resourceGroup}`);
 
-    const groupParameters = {
-      location: this.serverless.service.provider["location"]
-    };
-
-    return await this.resourceClient.resourceGroups.createOrUpdate(this.resourceGroup, groupParameters);
+    return await this.resourceClient.resourceGroups.createOrUpdate(this.resourceGroup, {
+      location: this.getRegion(),
```
mydiemho: rename `location` to `region` to be consistent

wbreza: Can't change it here - this is the call to the ARM sdk and it expects `location`.

src/shared/utils.ts
wbreza: @mydiemho initially added this in her PR.

src/services/baseService.ts
```diff
@@ -81,4 +87,33 @@ export abstract class BaseService {
   protected slsConfigFile(): string {
     return ("config" in this.options) ? this.options["config"] : "serverless.yml";
   }
+
+  private setDefaultValues(): void {
+    // TODO: Right now the serverless core will always default to AWS default region if the
+    // region has not been set in the serverless.yml or CLI options
+    const awsDefault = "us-east-1"
```
pjlittle: nit: feels odd to have the const within the function vs. more global -- both for the `awsDefault` and probably makes sense to have something like `azureDefault` that sits alongside it, so it's more clear what it's being used for.  Comment definitely helps.

src/services/functionAppService.ts
```diff
@@ -46,20 +47,27 @@ export class FunctionAppService extends BaseService {
...
 
     return await this.sendApiRequest("DELETE", deleteFunctionUrl);
   }
 
   public async syncTriggers(functionApp: Site) {
+    Guard.null(functionApp);
+
     this.serverless.cli.log("Syncing function triggers");
 
     const syncTriggersUrl = `${this.baseUrl}${functionApp.id}/syncfunctiontriggers?api-version=2016-08-01`;
```
pjlittle: Seeing the `api-version` everywhere makes me twitch a little bit, but probably more readable than trying to have some `buildUrl("syncfunctiontriggers") { return ...}` . And if it ain't broke...

src/services/baseService.ts
mydiemho: this won't be necessary anymore once serverless merge the other PR that remove the AWS default

src/services/functionAppService.ts
wbreza: There are many different versions and they are not consistent enough to warrant putting into constants right now.  Happy to do this later if needed.

src/services/baseService.ts
wbreza: We are only referencing these in this one function right now.  Agree if we needed to get the default values in other places we can bump this up.  Going to keep as is for the time being.

PR #175, opened 2019-06-20
util: add script to generate service principal to ease testing

mydiemho:
so user don't have to login every time they run `sls deploy`


scripts/generateSP.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+
+set -e
+
+SP_NAME=${1:-"http://FunctionsTestingPrincipal"}
+
+SUBSCRIPTION=$(az account show | jq .name)
```
pjlittle: super handy script :slightly_smiling_face: 

We should add some notes -- not everyone has jq available in their path (or knows what it is) -- we should just call that out.  Likewise for `${1:-"http://FunctionsTestingPrincipal"}` so that folks know you can pass in a value for the service principal name, if not will default to that Uri value.

Some comments at the top should suffice.

mydiemho: done

PR #176, opened 2019-06-21
feat: List deployments sub-command for deploy plugin

tbarlow12:
Resolves [AB#352]


src/plugins/deploy/azureDeployPlugin.ts
```diff
@@ -4,11 +4,41 @@ import { FunctionAppService } from "../../services/functionAppService";
...
 
   public constructor(private serverless: Serverless, private options: Serverless.Options) {
     this.hooks = {
-      "deploy:deploy": this.deploy.bind(this)
+      "deploy:deploy": this.deploy.bind(this),
+      "deploy:list:list": this.list.bind(this),
     };
+
+    this.commands = {
+      deploy: {
```
wbreza: Does defining these commands mess with the top level core plugin on the CLI ?

src/plugins/deploy/azureDeployPlugin.ts
```diff
@@ -4,11 +4,41 @@ import { FunctionAppService } from "../../services/functionAppService";
...
+      }
+    }
+  }
+
+  private async list() {
+    this.serverless.cli.log("Listing deployments");
+    const resourceService = new ResourceService(this.serverless, this.options);
+    const deployments = await resourceService.listDeployments();
+    let stringDeployments = "\n\nDeployments";
+    for (const dep of deployments) {
```
wbreza: If no deployments exist consider adding a line that says "No deployments found for resource group" or similar.

src/services/resourceService.ts
```diff
@@ -11,6 +11,11 @@ export class ResourceService extends BaseService {
     this.resourceClient = new ResourceManagementClient(this.credentials, this.subscriptionId);
   }
 
+  public async listDeployments() {
```
wbreza: This is more of a "get" operation and think we should rename to `getDeployments`.  The listing/writing is happening in the plugin.

src/plugins/deploy/azureDeployPlugin.ts
tbarlow12: I was concerned about that, which is why I left out the `lifecycleEvents` (I defined them at first, which caused `deploy` to be triggered twice). I set the plugin to log `this.commands` to see what the object looks like by default, and itame back undefined. In my manual testing, the deployment still works fine. Feel free to double check me on that though.

src/services/resourceService.test.ts
```diff
@@ -70,4 +72,15 @@ describe("Resource Service", () => {
     expect(ResourceManagementClient.prototype.resourceGroups.deleteMethod)
       .toBeCalledWith(resourceGroup);
   });
+
+  it("lists deployments", async () => {
```
mydiemho: another to test when deployments are empty

PR #178, opened 2019-06-25
fix: Ensures the correct SCM domain is found for uploading zip package

wbreza:
The SCM domain is different based on the type of function app deployment.  In a typical function app, the SCM domain is always {appName}.scm.azurewebsites.net.

When deploying a function into an App Service Environment (ASE) the SCM domain is a sub domain off of the ASE, ex) {appName}.scm.{asePath}.p.azurewebsites.net

This fix addresses the difference in finding the correct SCM domain to deploy the zip package.

### DIFF
--- a/src/services/functionAppService.test.ts
+++ b/src/services/functionAppService.test.ts
@@ -13,7 +13,6 @@ import { ArmDeployment, ArmTemplateType } from "../models/armTemplates";
 jest.mock("@azure/arm-resources")
 
 describe("Function App Service", () => {
-
   const app = MockFactory.createTestSite();
   const slsService = MockFactory.createTestService();
   const variables = MockFactory.createTestVariables();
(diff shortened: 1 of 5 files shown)


src/services/functionAppService.test.ts
```diff
@@ -197,11 +191,41 @@ describe("Function App Service", () => {
...
+  it("uploads functions with custom SCM domain (aka App service environments)", async () => {
+    const customApp = {
+      ...MockFactory.createTestSite("CustomAppWithinASE"),
+      enabledHostNames: [
+        "myapi.customase.p.azurewebsites.net",
+        "myapi.scm.customase.p.azurewebsites.net"
+      ],
+    }
+
+    const expectedUploadUrl = `https://${customApp.enabledHostNames[1]}/api/zipdeploy/`;
```
mydiemho: does Azure provide some API where we can check for this?  What if they change the scm domain url?

wbreza: @mydiemho Not that i am aware of today.  Looks like these are the 2 well known varients.
https://blogs.msdn.microsoft.com/benjaminperkins/2017/11/08/how-to-access-kudu-scm-for-an-azure-app-service-environment-ase/

pjlittle: heh. fair enough.

mydiemho: can we include this as a comment so future developers know where this came from?

PR #179, opened 2019-06-25
fix: Simple fix in test of equality comparison

tbarlow12:
Discovered this small bug in the test. Overloading the variable name `masterKey`, using `result` instead.

### DIFF
--- a/src/services/functionAppService.test.ts
+++ b/src/services/functionAppService.test.ts
@@ -107,8 +107,8 @@ describe("Function App Service", () => {
 
   it("gets master key", async () => {
     const service = createService();
-    const masterKey = await service.getMasterKey();
-    expect(masterKey).toEqual(masterKey);
+    const result = await service.getMasterKey();
+    expect(result).toEqual(masterKey);
 
   });
 


src/services/functionAppService.test.ts
```diff
@@ -107,8 +107,8 @@ describe("Function App Service", () => {
 
   it("gets master key", async () => {
```
mydiemho: suggestion: rename test description to start with `should/should not`

src/services/functionAppService.test.ts
```diff
@@ -107,8 +107,8 @@ describe("Function App Service", () => {
...
     const service = createService();
-    const masterKey = await service.getMasterKey();
-    expect(masterKey).toEqual(masterKey);
+    const result = await service.getMasterKey();
```
mydiemho: nit: I would call this `actualMasterKey`

src/services/functionAppService.test.ts
tbarlow12: will do

src/services/functionAppService.test.ts
tbarlow12: will do

PR #180, opened 2019-06-26
Fixes issue where compile errors are not caught during build steps

wbreza:
- Fixes missing import statements in MockFactory
- Adds new CI step to confirm TypeScript compile succeeds before running tests.


.travis.yml
```diff
@@ -26,17 +28,23 @@ install:
 
 jobs:
   include:
+    - stage: compile
+      name: "TypeScript compiler for Node"
+      node_js:
```
mydiemho: this is not running `node 10` anymore

.travis.yml
```diff
@@ -26,17 +28,23 @@ install:
 
 jobs:
   include:
+    - stage: compile
+      name: "TypeScript compiler for Node"
+      node_js: 
+        - "8"
+        - "10"
+      script: 
+        - npm run compile
```
mydiemho: do you just want to compile before running unit tests?

.travis.yml
wbreza: What is the correct syntax to run on both?

wbreza: Figured it out - reverted most of what i had and just added the compile in the existing stage.

wbreza:
@mydiemho Please take a second look. Thanks!

package.json
```diff
@@ -14,8 +14,9 @@
     "test": "jest",
     "test:ci": "npm run test -- --ci",
     "test:coverage": "npm run test -- --coverage",
+    "compile": "tsc",
     "prebuild": "shx rm -rf lib/ && npm run test",
-    "build": "tsc"
+    "build": "npm run compile"
```
mydiemho: is this just for backward compatibility?  seem redundant

PR #183, opened 2019-06-27
 feat: Updates to deployment process to enable rollback (Part 1)

tbarlow12:
Updating the deployment process will take a few pull requests to keep them (relatively) small. Here are the changes included in this pull request.
- Update default naming convention of Azure Storage account (and a utility function that assists with creating name)
- Stubs for rollback enabled (currently switched off to avoid breaking tests, will switch on in future PR), which includes
    - Timestamping names of deployments
    - Timestamping names of packages
    - Initialization of deployment artifact container
- Updates the `LoginService` to accept options, which is used in the `AzureBlobStorageService` to retrieve a storage-specific access token
- Remove dead code from `FunctionAppService`

Resolves [AB#357], [AB#408], [AB#414], [AB#415]


src/services/functionAppService.ts
```diff
@@ -201,29 +213,6 @@ export class FunctionAppService extends BaseService {
     };
   }
 
-  private async runKuduCommand(functionApp: Site, command: string) {
```
tbarlow12: This is not being used anywhere

docs/DEPLOY.md
```diff
@@ -56,3 +56,48 @@ then user try to deploy
...
+plugins:
+  - serverless-azure-functions
+
+package:
+  ...
+
+deploy:
+  # Rollback enabled, deploying to blob storage
+  # Default is true
+  # If false, deploys directly to function app
```
pjlittle: nice, i like the flexibility :+1:

src/config.ts
```diff
@@ -10,10 +11,11 @@ export const constants = {
   logStreamApiPath: "/api/logstream/application/functions/function/",
   masterKeyApiPath: "/api/functions/admin/masterkey",
   providerName: "azure",
+  rollbackEnabled: false,
```
pjlittle: the README says the default is true -- whichever we choose, just need to be consistent.

src/services/azureBlobStorageService.ts
```diff
@@ -149,4 +155,12 @@ export class AzureBlobStorageService extends BaseService {
       blobName,
     );
   }
+
+  private async getToken(): Promise<string> {
+    const authResponse = await AzureLoginService.login({
+      tokenAudience: "https://storage.azure.com/"
```
pjlittle: if this is only used here, I'm good, otherwise let's promote to a const somewhere.

src/config.ts
tbarlow12: Right, the default in this PR is false just to not break everything. The next PR will set this to true

PR #184, opened 2019-06-28
fix: Add sequence diagram png files

tbarlow12:
[Github does not yet render Mermaid diagrams](https://github.com/github/markup/issues/533). As a workaround, this includes snapshots of the rendered diagram and links to the images


docs/sequenceDiagrams/deployBlob.md
```diff
@@ -0,0 +1,17 @@
...
+  participant f as Function App
+  participant b as Blob Storage
+
+  note right of s: `sls deploy`
+  s ->> r: Create resource group
+  s ->> r: Deploy ARM template
+  r ->> f: Included in ARM template
+  r ->> b: Included in ARM template
+  note right of s: Zip code
+  s ->> b: Deploy zip code with name {appName}-v{version}.zip
```
pjlittle: As per current design, this should include timestamp/commit SHA, instead of version.  It can stay if we want -- just need a more unique value as well.

PR #185, opened 2019-06-28
feat: Updates to deployment process to enable rollback (Part 2)

tbarlow12:
## High Level Overview
- Uploads a timestamped archive of code to Azure Blob Storage
- Adds same timestamp to ARM deployments to enable linking
- **Does not set `WEBSITE_RUN_FROM_PACKAGE` to use package directly from blob storage**
  - Currently just used as an archive. Initially, thought to update the `WEBSITE_RUN_FROM_PACKAGE` setting with the URL for the blob (hence the implementation of the SAS URL generator), but discovered [issues with cold start when using that methodology to deploy to Windows](https://docs.microsoft.com/en-us/azure/azure-functions/run-functions-from-deployment-package#enabling-functions-to-run-from-a-package).
  - Future `rollback` implementation will download the appropriate archive and re-deploy it to the Function App.
  - This may (and probably will) change as the Azure Functions Premium plan is released on Linux or cold start issues are mitigated

## Details

- Adds dependency on the `@azure/arm-storage` package
- Sets `rollbackEnabled` as default
- Updates broken tests because of timestamped deployment name
- Add tests for blob storage operations
- Allows user to still specify deployment name with rollback

Resolves [AB#388], [AB#358], [AB#414] and [AB#415]


src/services/functionAppService.ts
```diff
@@ -43,22 +51,11 @@ export class FunctionAppService extends BaseService {
     return response.data.value;
   }
 
-  /**
-   * Initialize deployment artifact container if rollback is specified
-   */
-  public async initialize() {
```
PIC123: Is this unnecessary because of the added params in the constructor?

tbarlow12: Rather than initializing the blob storage here, it initializes before actually creating the blob. The problem was it was trying to initialize and create a container for an account that didn't yet exist

src/services/functionAppService.ts
```diff
@@ -173,25 +161,58 @@ export class FunctionAppService extends BaseService {
...
+          this.log(`-> ${functionConfig.name}: ${method} ${httpConfig.url}`);
         }
       }
     });
   }
 
+  /**
+   * Uploads artifact file to blob storage container 
+   */
+  private async uploadFunctionAppToBlobStorage() {
```
mydiemho: I would call this `uploadZipArtifactToBlobStorage` or `uploadFunctionArtifactToBlobStorage`

src/services/functionAppService.ts
```diff
@@ -124,6 +117,7 @@ export class FunctionAppService extends BaseService {
     this.log("Deploying serverless functions...");    
 
     await this.zipDeploy(functionApp);
+    await this.uploadFunctionAppToBlobStorage();
```
mydiemho: commented suggestion below but this name is kinda misleading, both are zip deploy, the only difference is zip storage location

mydiemho: also why are we uploading to both place? what is the value of `RUN_FROM_PACKAGE`?

PR #188, opened 2019-07-03
Feat: Invoke Plugin

njm3754:
Invoke Plugin added
NOTE: Use the dev branch for deployment of functions. This branch has older version of deploy.
Example of how to invoke a function by passing data from command line.
```bash
sls invoke -f {functionName} --data '{"name": "test"}'
```
Example of how to invoke a function by passing the file path from command line.
Note: The file containing the data should be present in the project directory (./project/data.json)

```bash
sls invoke -f {functionName} --path data.json
```
Resolves [AB#309]


src/plugins/invoke/azureInvoke.test.ts
```diff
@@ -0,0 +1,98 @@
...
+    options["method"] = "GET";
+
+    const plugin = new AzureInvoke(sls, options);
+    await invokeHook(plugin, "invoke:invoke");
+    expect(invoke).toBeCalledWith(options["function"], fileContent, options["method"]);
+    expect(sls.cli.log).toBeCalledWith(expectedResult.data);
+  
+  });
+
+  it("calls the invooke hook with file path", async () => {
```
wbreza: Spelling "invoke"

src/plugins/invoke/azureInvoke.ts
```diff
@@ -1,41 +1,79 @@
...
+          path: {
+            usage: "Path to file to put in body",
+            shortcut: "p"
+          },
+          data: {
+            usage: "Data string for body of request",
+            shortcut: "d"
+          },
+          method: {
+            usage: "GET or POST request (Default is GET)",
```
wbreza: Do we support other HTTP verbs besides GET and POST?

src/plugins/invoke/azureInvoke.ts
```diff
@@ -1,41 +1,79 @@
...
+    const functionName = this.options["function"];
+    const data = this.options["data"];
+    const method = this.options["method"] || "GET";
+    if (!functionName) {
+      this.serverless.cli.log("Need to provide a name of function to invoke");
+      return;
     }
-
-    return this.provider.invoke(func, eventType, this.options["data"]);
+    if (!data) {
```
wbreza: Should data be required?  If I have a function that is a GET that doesn't expect any params or body I would expect it to still work.

src/services/invokeService.ts
```diff
@@ -0,0 +1,86 @@
+import { BaseService } from "./baseService"
+import Serverless from "serverless";
+import config from "../config";
+import axios from "axios";
+import { FunctionAppService } from "./functionAppService";
+import { FunctionAppResource } from "../armTemplates/resources/functionApp";
+import { ServerlessAzureConfig } from "../models/serverless";
+
+export class InvokeService extends BaseService {
+  public serverless: Serverless;
```
wbreza: These are not required - when using the Base service the serverless and options are already made available via a protected member by calling `super(...)` method.

src/services/invokeService.ts
```diff
@@ -0,0 +1,86 @@
...
+  public options: Serverless.Options;
+  public constructor(serverless: Serverless, options: Serverless.Options) {
+    super(serverless, options);
+    this.serverless = serverless;
```
wbreza: Not required - remove.

src/services/invokeService.ts
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  /**
+   * Invoke an Azure Function
+   * @param functionName Name of function to invoke
+   * @param data Data to use as body or query params
+   * @param method GET or POST
+   */
+  public async invoke(functionName: string, data: any, method: string){
+    /* accesses the admin key */
```
wbreza: Comment doesn't make sense with the code.

src/services/invokeService.ts
```diff
@@ -0,0 +1,86 @@
...
+    this.options = options;
+  }
+
+  /**
+   * Invoke an Azure Function
+   * @param functionName Name of function to invoke
+   * @param data Data to use as body or query params
+   * @param method GET or POST
+   */
+  public async invoke(functionName: string, data: any, method: string){
```
wbreza: Consider reordering your params to method, functionName, then data.  Data seems like it should also be optional since not all calls will require data.

src/services/invokeService.ts
```diff
@@ -0,0 +1,86 @@
...
+    /* accesses the admin key */
+    if (!(functionName in this.slsFunctions())) {
+      this.serverless.cli.log(`Function ${functionName} does not exist`);
+      return;
+    }
+    const functionObject = this.slsFunctions()[functionName];
```
wbreza: You are already making call to `slsFunctions()` above.  Consider caching that result and reusing here to avoid the duplicate call.

src/services/invokeService.ts
```diff
@@ -0,0 +1,86 @@
...
+    const eventType = Object.keys(functionObject["events"][0])[0];
+    if (eventType !== "http") {
+      this.log("Needs to be an http function");
+      return;
+    }
+
+    let url = `http://${FunctionAppResource.getResourceName(this.serverless.service as any)}${config.functionAppDomain}${config.functionAppApiPath + functionName}`;
```
wbreza: `config.functionAppDomain` from the constants is not correct in all cases.  I would use the function app service to get to domain from the actual function app configuration.  Also - I would remove dependency on the `config` constants all together since in this updated model they are not really used any more.  Take a look @ `FunctionAppService.getFunctionHttpTriggerConfig` for more info.

The `functionName` may not always be used as the api endpoint.  This is configurable as part of the binding.

src/services/invokeService.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    if (method === "GET") {
```
wbreza: This will fail when `data` isn't defined.  We need to support GET requests without data or querystring.

src/services/invokeService.ts
```diff
@@ -0,0 +1,86 @@
...
+   * Get options object  
+   * @param method The method used (POST or GET)
+   * @param data Data to use as body or query params
+   */
+  private async getOptions(method: string, data?: any) {
+    
+    const functionAppService = new FunctionAppService(this.serverless, this.options);
+    const functionsAdminKey = await functionAppService.getMasterKey();
+    const options: any = {
+      host: config.functionAppDomain,
```
wbreza: Same as above - You'll need to retrieve the function app domain from the function app dynamically.

src/services/invokeService.ts
```diff
@@ -0,0 +1,86 @@
...
+      headers: {
+        "x-functions-key": functionsAdminKey
+      },
+      method,
+    };
+    if (method === "POST" && data) {
```
wbreza: I believe there is a `data` on the axios options that you can set vs manually setting it on the body.

wbreza: https://github.com/axios/axios#axios-api

src/plugins/invoke/azureInvoke.ts
```diff
@@ -1,41 +1,79 @@
...
+        usage: "Invoke command",
+        lifecycleEvents: [
+          "invoke"
+        ],
+        options: {
+          function: {
+            usage: "Function to call",
+            shortcut: "f",
+          },
+          path: {
```
wbreza: I see `path` is an option - but i don't see it being using within the invoke service.

src/plugins/invoke/azureInvoke.ts
```diff
@@ -1,41 +1,79 @@
...
     }
-
-    return this.provider.invoke(func, eventType, this.options["data"]);
+    if (!data) {
+      this.serverless.cli.log("Need to provide data or path");
+      return;
+    }
+    this.invokeService = new InvokeService(this.serverless, this.options);
+    const response =  await this.invokeService.invoke(functionName, data, method);
+    this.serverless.cli.log(response.data);
```
wbreza: This will likely be a JSON response.  If so we should call `JSON.stringify(...)` to make it display correctly.

src/plugins/invoke/azureInvoke.test.ts
```diff
@@ -0,0 +1,98 @@
...
+    InvokeService.prototype.invoke = invoke;
+    const sls = MockFactory.createTestServerless();
+    const options = MockFactory.createTestServerlessOptions();
+    options["function"] = "testApp";
+    options["path"] = "garbage.json";
+    options["method"] = "GET";
+    expect(() => new AzureInvoke(sls, options)).toThrow();
+  });
+
+  it("The invoke function fails when no data is passsed", async () => {
```
wbreza: This shouldn't fail.  `data` should be optional.

src/plugins/invoke/azureInvoke.ts
```diff
@@ -1,41 +1,79 @@
...
         : join(this.serverless.config.servicePath, path);
+      this.serverless.cli.log(this.serverless.config.servicePath);
+      this.serverless.cli.log(path);
 
-      if (!this.serverless.utils.fileExistsSync(absolutePath)) {
+      if (!fs.existsSync(absolutePath)) {
         throw new Error("The file you provided does not exist.");
       }
-      this.options["data"] = this.serverless.utils.readFileSync(absolutePath);
+      this.options["data"] = fs.readFileSync(absolutePath).toString();
```
wbreza: I now see that you are setting `data` to be the file contents... Ignore comments below regarding non usage of `path`.

src/plugins/invoke/azureInvoke.ts
```diff
@@ -1,41 +1,79 @@
...
+import fs from "fs";
 
 export class AzureInvoke {
   public hooks: { [eventName: string]: Promise<any> };
+  private commands: any;
   private provider: AzureProvider;
-
+  private login: AzureLoginPlugin
+  private invokeService: InvokeService;
   public constructor(private serverless: Serverless, private options: Serverless.Options) {
```
wbreza: Consider creating a new interface for `InvokeOptions` so we can make this more strongly typed and avoid using the square braket syntax.

src/plugins/invoke/azureInvoke.ts
```diff
@@ -1,41 +1,79 @@
...
+import { InvokeService } from "../../services/invokeService";
+import { AzureLoginPlugin } from "../login/loginPlugin";
+import fs from "fs";
 
 export class AzureInvoke {
   public hooks: { [eventName: string]: Promise<any> };
+  private commands: any;
   private provider: AzureProvider;
-
+  private login: AzureLoginPlugin
```
wbreza: Do we need a reference to the login plugin? I would imagine the login plugin would be adjusted to login during the invoke process.

src/plugins/invoke/azureInvoke.ts
```diff
@@ -1,41 +1,79 @@
...
+  private invokeService: InvokeService;
   public constructor(private serverless: Serverless, private options: Serverless.Options) {
     this.provider = (this.serverless.getProvider("azure") as any) as AzureProvider;
```
wbreza: `this.provider` should no longer be used and can be removed.

src/plugins/invoke/azureInvoke.test.ts
```diff
@@ -0,0 +1,98 @@
+import { MockFactory } from "../../test/mockFactory";
+import { invokeHook } from "../../test/utils";
+import mockFs from "mock-fs";
+import { AzureInvoke } from "./azureInvoke";
+import { InvokeService } from "../../services/invokeService";
```
tbarlow12: Import `InvokeService` after mocking the `invokeService` module

src/plugins/invoke/azureInvoke.test.ts
tbarlow12: Yup, in a scenario where you're just hitting an API that requires no body or query params, you wouldn't need data.

src/services/invokeService.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    // Master Key
+    axiosMock.onGet(masterKeyUrl).reply(200, { value: masterKey });
+    // Auth Key
+    axiosMock.onGet(authKeyUrl).reply(200, authKey);
+    axiosMock.onGet(functionUrl).reply(200, result);
+  });
+  beforeEach(() => {
+    FunctionAppService.prototype.getMasterKey = jest.fn();
+  });
+
```
tbarlow12: nit: Extra new line

bash.exe.stackdump
```diff
@@ -0,0 +1,16 @@
+Stack trace:
```
tbarlow12: Remove this file

src/services/invokeService.test.ts
```diff
@@ -0,0 +1,73 @@
...
+  const authKey = "authKey";
+  const baseUrl = "https://management.azure.com"
+  const masterKeyUrl = `https://${app.defaultHostName}/admin/host/systemkeys/_master`;
+  const authKeyUrl = `${baseUrl}${app.id}/functions/admin/token?api-version=2016-08-01`;
+  let urlGET = `http://${app.defaultHostName}/api/hello?name=${testData}`;
+  let urlPOST = `http://${app.defaultHostName}/api/hello`;
+  let masterKey: string;
+
+  beforeAll(() => {
+
```
tbarlow12: nit: remove blank lines 24 and 34

src/services/invokeService.ts
```diff
@@ -0,0 +1,87 @@
...
+import Serverless from "serverless";
+import axios from "axios";
+import { FunctionAppService } from "./functionAppService";
+
+export class InvokeService extends BaseService {
+  public functionAppService: FunctionAppService;
+  public constructor(serverless: Serverless, options: Serverless.Options) {
+    super(serverless, options);
+    this.functionAppService = new FunctionAppService(serverless, options);
+
```
tbarlow12: nit: remove blank line on 11, add line before 8

tbarlow12:
@njm3754 it looks like you have some unsigned commits and then some conflicts

src/services/invokeService.ts
wbreza: This still calls the function 2x.  Consider caching the result of the function call and resusing the variable.

src/services/invokeService.ts
```diff
@@ -0,0 +1,87 @@
...
+      try {
+        eventData = JSON.parse(eventData);
+      }
+      catch (error) {
+        return Promise.reject("The specified input data isn't a valid JSON string. " +
+        "Please correct it and try invoking the function again.");
+      }
+    }
+    return Object.keys(eventData)
+      .map((key) => `${key}=${eventData[key]}`)
```
wbreza: You should urlencode the right side of the query string expression.

PR #189, opened 2019-07-03
feat: add support for specifying resourceGroup in both cli and yaml

mydiemho:
when a name is specified in cli or yaml, serverless will use this name to create the resourceGroup instead of creating a name from convention. 

all underlying resources will still be created with a convention name.

### DIFF
--- a/src/plugins/deploy/azureDeployPlugin.ts
+++ b/src/plugins/deploy/azureDeployPlugin.ts
@@ -24,7 +24,7 @@ export class AzureDeployPlugin {
         },
         options: {
           "resourceGroup": {
-            usage: "resource group for the service",
+            usage: "Resource group for the service",
             shortcut: "g",
           }
         }
@@ -32,16 +32,12 @@ export class AzureDeployPlugin {
     }
   }
 
-  private log(message: string) {
-    this.serverless.cli.log(message);
-  }
-
   private async list() {
     this.serverless.cli.log("Listing deployments");
     const resourceService = new ResourceService(this.serverless, this.options);
     const deployments = await resourceService.getDeployments();
     if (!deployments || deployments.length === 0) {
-      this.log(`No deployments found for resource group '${resourceService.getResourceGroupName()}'`);
+      this.serverless.cli.log(`No deployments found for resource group '${resourceService.getResourceGroupName()}'`);
       return;
     }
     let stringDeployments = "\n\nDeployments";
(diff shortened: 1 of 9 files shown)


src/plugins/deploy/azureDeployPlugin.ts
```diff
@@ -21,17 +21,27 @@ export class AzureDeployPlugin {
               "list"
             ]
           }
+        },
+        options: {
+          "resourceGroup": {
+            usage: "resource group for the service",
```
wbreza: Capital "R" in "Resource Group" in usage param.

src/plugins/deploy/azureDeployPlugin.ts
```diff
@@ -21,17 +21,27 @@ export class AzureDeployPlugin {
...
+            shortcut: "g",
+          }
         }
       }
     }
   }
 
+  private log(message: string) {
```
wbreza: Instead of introducing this here consider creating some base `ServerlessPlugin` class that all our plugins inherit from to provide some base functionality like this `log(...)` function..

src/services/baseService.test.ts
```diff
@@ -84,48 +85,60 @@ describe("Base Service", () => {
...
-  it("Sets default region and stage values if not defined", () => {
-    const mockService = new MockService(sls);
-
-    expect(mockService).not.toBeNull();
-    expect(sls.service.provider.region).toEqual("westus");
-    expect(sls.service.provider.stage).toEqual("dev");
-  });
-
-  it("returns region and stage based on CLI options", () => {
+  it("use the resource group name specified in CLI", () => {
```
wbreza: Consider renaming the test description to be inclusive of region and stage values in addition to the resource group.

src/services/baseService.ts
```diff
@@ -46,15 +46,21 @@ export abstract class BaseService {
...
   }
 
+  public getPrefix(): string {
+    return this.config.provider.prefix;
+  }
+
   public getResourceGroupName(): string {
-    return this.options["resourceGroup"]
+    const name = this.options.resourceGroup
       || this.serverless.service.provider["resourceGroup"]
```
wbreza: Consider changing to `this.config.provider.resourceGroup` since the `config` property is now scoped to the `ServerlessAzureConfig` interface.

src/plugins/deploy/azureDeployPlugin.ts
mydiemho: good idea.  I'm going to undo this and do the changes in a separate pr

src/services/baseService.test.ts
mydiemho: there's already another test for region and stage.  I just thought separating the test make more sense because we don't always specific the region, stage, and resourceGroup flag together.  Thoughts?

wbreza: OK - from the diff it looks like you are replacing the existing text.

PR #190, opened 2019-07-04
fix: Update storage account naming convention

wbreza:
Ensures storage account names do not clash based on combination of prefix, region, stage & service name

- [x] Ensures generated name is <= 24 chars
- [x] Ensures region names don't conflict
- [x] Generates short region part from full value
- [x] Generates short stage part from full value
- [x] Includes partial hash of service name in full name

Resolves AB#420


src/shared/utils.ts
```diff
@@ -136,4 +137,61 @@ export class Utils {
...
+  /**
+   * Creates a short name from a long name based on a well-known string map
+   * @param longName The long name to replace
+   * @param wellKnownMap A well known map of long terms to short abbreviations
+   */
+  private static createShortName(longName: string, wellKnownMap: { [key: string]: string }) {
+    Guard.empty(longName);
+    Guard.null(wellKnownMap);
+
+    const pattern = `(${Object.keys(wellKnownMap).join("|")})`;
```
mydiemho: it would be good to have comments explaining what these pattern and regex do

PR #191, opened 2019-07-04
feat: Rollback Plugin

tbarlow12:
## Updates
- Adds new `rollback` plugin
- Fixes `deploy list` to show actual timestamp from name, which is used for rollback
- Adds required functionality to `AzureBlobStorageService` and `ArmService`
- Adds previously missing tests of required functionality

## How to Use
1. Get a timestamp. Use
```
sls deploy list
```
to retrieve one.
2. Roll back function app to timestamp
```
sls rollback -t <timestamp>
```

## Options
In `deploy` section of `serverless.yml`, you can specify whether to run from a blob URL or deploy the package directly to the function app by setting:
```yaml
deploy:
  # Default is false
  runFromBlobUrl: true|false
```

Resolves [AB#317]

### DIFF
--- a/src/plugins/deploy/azureDeployPlugin.test.ts
+++ b/src/plugins/deploy/azureDeployPlugin.test.ts
@@ -36,7 +36,29 @@ describe("Deploy plugin", () => {
     expect(uploadFunctions).toBeCalledWith(functionAppStub);
   });
 
-  it("lists deployments", async () => {
+  it("lists deployments with timestamps", async () => {
+    const deployments = MockFactory.createTestDeployments(5, true);
+    ResourceService.prototype.getDeployments = jest.fn(() => Promise.resolve(deployments));
+    const sls = MockFactory.createTestServerless();
+    const options = MockFactory.createTestServerlessOptions();
+    const plugin = new AzureDeployPlugin(sls, options);
+    await invokeHook(plugin, "deploy:list:list");
+    let expectedLogStatement = "\n\nDeployments";
+    const originalTimestamp = +MockFactory.createTestTimestamp();
+    let i = 0
+    for (const dep of deployments) {
+      const timestamp = originalTimestamp + i
+      expectedLogStatement += "\n-----------\n"
+      expectedLogStatement += `Name: ${dep.name}\n`
+      expectedLogStatement += `Timestamp: ${timestamp}\n`;
+      expectedLogStatement += `Datetime: ${new Date(timestamp).toISOString()}\n`
+      i++
+    }
+    expectedLogStatement += "-----------\n"
+    expect(sls.cli.log).lastCalledWith(expectedLogStatement);
+  });
+
+  it("lists deployments without timestamps", async () => {
     const deployments = MockFactory.createTestDeployments();
     ResourceService.prototype.getDeployments = jest.fn(() => Promise.resolve(deployments));
     const sls = MockFactory.createTestServerless();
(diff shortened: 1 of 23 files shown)


src/services/armService.ts
```diff
@@ -97,7 +97,7 @@ export class ArmService extends BaseService {
     Object.keys(deployment.parameters).forEach((key) => {
       const parameterValue = deployment.parameters[key];
       if (parameterValue) {
-        deploymentParameters[key] = { value: deployment.parameters[key] };
+        deploymentParameters[key] = (typeof parameterValue === "string") ? { value: parameterValue } : { value: parameterValue.value }
```
tbarlow12: @wbreza When I get the ARM template from the previous deployment, the parameter values look like:
```
{
   type: "String",
   value: "value1"
}
```
When I just used that object rather than extracting the value, it fails. I guess it doesn't like the `type` attribute for some reason. Do you have any cleaner ways for how to do this?

src/plugins/deploy/azureDeployPlugin.test.ts
```diff
@@ -36,7 +36,29 @@ describe("Deploy plugin", () => {
...
+      expectedLogStatement += `Name: ${dep.name}\n`
+      expectedLogStatement += `Timestamp: ${timestamp}\n`;
+      expectedLogStatement += `Datetime: ${new Date(timestamp).toISOString()}\n`
+      i++
+    }
+    expectedLogStatement += "-----------\n"
+    expect(sls.cli.log).lastCalledWith(expectedLogStatement);
+  });
+
+  it("lists deployments without timestamps", async () => {
```
wbreza: Why would a deployment NOT have a timestamp?

wbreza: I think i see now - You are interpolating it from the file name - not the timestamp from the actual deployment?

src/services/armService.ts
wbreza: The `deployTemplate(...)` is called it is called with deployment params of simple key/value pairs and then converted into the final format ARM templates expect by default.  What I imagine is happening is you are passing the params as already final constructed ARM params from the pre-existing deployment.

src/services/functionAppService.ts
```diff
@@ -116,8 +114,9 @@ export class FunctionAppService extends BaseService {
 
     this.log("Deploying serverless functions...");    
 
-    await this.uploadZippedArfifactToFunctionApp(functionApp);
-    await this.uploadZippedArtifactToBlobStorage();
+    const functionZipFile = this.getFunctionZipFile();
+    await this.uploadZippedArfifactToFunctionApp(functionApp, functionZipFile);
```
wbreza: Can these be done in parallel via `Promise.all()` ??

src/services/functionAppService.ts
```diff
@@ -138,16 +137,16 @@ export class FunctionAppService extends BaseService {
     return await this.get();
   }
 
-  private async uploadZippedArfifactToFunctionApp(functionApp) {
+  public async uploadZippedArfifactToFunctionApp(functionApp: Site, functionZipFile: string) {
```
wbreza: why switching to `public`? Needed?

src/services/rollbackService.ts
```diff
@@ -0,0 +1,146 @@
...
+  /**
+   * Convert previous deployment to ArmDeployment
+   * @param deployment Previous deployment retrieved from Azure SDK
+   */
+  private async convertToArmDeployment(deployment: DeploymentExtended): Promise<ArmDeployment> {
+    const resourceService = new ResourceService(this.serverless, this.options);
+    const { template } = await resourceService.getDeploymentTemplate(deployment.name);
+    return {
+      template,
+      parameters: deployment.properties.parameters,
```
wbreza: If you convert the params back to the expected format here you can remove the special case in the `ArmService` that you had asked me about.

src/services/functionAppService.ts
tbarlow12: It's used inside the rollback service now. to re-deploy function zip to the function app

src/services/rollbackService.ts
tbarlow12: will do. thanks

src/plugins/deploy/azureDeployPlugin.test.ts
tbarlow12: I "stamp" it when the package is being created. I wanted to use the timestamp from the deployment, but since someone "could" have the same deployment name, I'd have to scan through all deployments to check which one was most recent. It seemed more safe to just include the timestamp in the name of the deployment. I checked the response from the deployment of the resource group, and the only property included is the provisioning state

PR #192, opened 2019-07-05
fix: Exit process when error logging into Azure

tbarlow12:


src/plugins/login/loginPlugin.ts
```diff
@@ -33,6 +33,7 @@ export class AzureLoginPlugin {
     catch (e) {
       this.serverless.cli.log("Error logging into azure");
       this.serverless.cli.log(`${e}`);
+      process.exit(0);
```
wbreza: Can we add a unit test to validate this behavior?

PR #193, opened 2019-07-05
refactor: add base plugin and rename plugins to follow convention

mydiemho:
- [x]  All plugin classes now extends from base plugin
- [x] All plugin classes rename to follow naming convention


.gitignore
```diff
@@ -2,4 +2,5 @@ node_modules/
 /lib/
 *.dll
 /coverage/
-.env*
\ No newline at end of file
+.env*
+.idea/
```
wbreza: What is `.idea` ??

tbarlow12: that sounds like intellij stuff

tbarlow12: or probably webstorm I'm guessing

tbarlow12:
How did coverage decrease 10%?

.gitignore
mydiemho: yup intellij/webstorm metadata

mydiemho:
> How did coverage decrease 10%?

looks like a few plugins didn't have enough tests, and renaming them cause them to be counted

PR #201, opened 2019-07-12
docs: README Update

tbarlow12:
Update README to reflect current state of plugin.

[AB#483]

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,9 +201,15 @@ We're still in the process of getting everying running 100%, but please refer to
 1. Clone this repository to your local machine
 2. Navigate to the cloned folder
 3. Run `npm install`
-4. Run `npm run-script build`
+4. Run `npm run build`
 5. Navigate to a folder where you created a new Serverless project, run `npm install`, and then run `npm link {path to serverless-azure-functions folder}`. Running `npm install` after the link command may override the link.
 6. The npm modules should now contain your local version of this plugin.
+
+#### Unit Tests
+
+We use [Jest](https://jestjs.io/) for unit tests, and it is expected that every Pull Request containing code changes have accompanying unit tests.
+
+Run unit tests using `npm test` or `npm run test:coverage` to get coverage results.
 
 #### Signing commits
 
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
-# Azure Functions Serverless Plugin 
+# Azure Functions Serverless Plugin
 
 This plugin enables Azure Functions support within the Serverless Framework.
 
 [![Coverage Status](https://coveralls.io/repos/github/serverless/serverless-azure-functions/badge.svg)](https://coveralls.io/github/serverless/serverless-azure-functions)
 
-## Getting started
+## Quickstart
 
 ### Pre-requisites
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,61 +1,46 @@
...
 [![Coverage Status](https://coveralls.io/repos/github/serverless/serverless-azure-functions/badge.svg)](https://coveralls.io/github/serverless/serverless-azure-functions)
 
-## Getting started
+## Quickstart
 
 ### Pre-requisites
 
 1. Node.js v6.5.0+ *(this is the runtime version supported by Azure Functions)*
 2. Serverless CLI `v1.9.0+`. You can run `npm i -g serverless` if you don't already have it.
 3. An Azure account. If you don't already have one, you can sign up for a [free trial](https://azure.microsoft.com/en-us/free/) that includes $200 of free credit.
```
ahemavathy: 4. Azure plugin that allows you to work with Azure Functions 
```
npm install -g serverless-azure-functions
```

README.md
```diff
@@ -1,61 +1,46 @@
...
-
-### Creating or removing Azure Functions
-
-To create a new Azure Function within your function app, run the following command from within your app's directory:
+### Create a new Azure Function App
 
 ```bash
-sls func add -n {functionName}
+# Create Azure Function App from template
+$ sls create -t azure-nodejs -p <appName>
```
ahemavathy: The sample app created with the template has its own readme.md file which is outdated. it needs to have the same contents as this document. Could you please modify that as well

README.md
ahemavathy: Azure plugin that allows you to work with Azure Functions
npm install -g serverless-azure-functions

README.md
tbarlow12: Sure thing. That will be the next step, updating both that README as well as the quickstart docs in the Serverless core

README.md
```diff
@@ -65,58 +50,146 @@ To clean up files generated from the build, you can simply run:
...
 
-### Advanced Authentication
+### Roll Back Your Function App
+
+To roll back your function app to a previous deployment, simply select a timestamp of a previous deployment and use `rollback` command.
+
+```bash
+# List all deployments to know the timestamp for rollback
+$ sls deploy list
+Serverless:
```
ahemavathy: this command doesn't seem to work for me - I have done multiple deployments but it returns nothing

README.md
```diff
@@ -65,58 +50,146 @@ To clean up files generated from the build, you can simply run:
...
-$env:azureServicePrincipalTenantId='<tenantId>'
-$env:azureServicePrincipalClientId='<servicePrincipalName>'
-$env:azureServicePrincipalPassword='<password>'
+### Creating or removing Azure Functions
+
+To create a new Azure Function within your function app, run the following command from within your app's directory:
+
+```bash
+# -n or --name for name of new function
+$ sls func add -n {functionName}
```
ahemavathy: this doesn't work either for me - command func not found

tbarlow12: @ahemavathy This makes me think you might be working off of an outdated version. Are you using npm link? Where did you install the plugin from?

README.md
```diff
@@ -65,58 +50,146 @@ To clean up files generated from the build, you can simply run:
...
 ```
 
+This will create a new handler file at the root of your project with the title `{functionName}.js`. It will also update `serverless.yml` to contain the new function.
+
+To remove an existing Azure Function from your function app, run the following command from within your app's directory:
+
+```bash
+# -n or --name for name of function to remove
+$ sls func remove -n {functionName}
```
ahemavathy: ditto

README.md
```diff
@@ -1,61 +1,46 @@
...
-npm i azure-functions-core-tools -g
+$ npm i azure-functions-core-tools -g
 ```
 
 Then, at the root of your project directory, run:
 
 ```bash
 # Builds necessary function bindings files
-sls offline
+$ sls offline
```
ahemavathy: this isin't present either

README.md
```diff
@@ -1,61 +1,46 @@
-# Azure Functions Serverless Plugin 
+# Azure Functions Serverless Plugin
```
wbreza: We'll need to add more in addition to the quickstart.  But this is good for now.

pjlittle:
Did we ever resolve Hema's comments?

README.md
tbarlow12: Because that's in the `devDependencies`, the `npm install` should take care of it.

README.md
```diff
@@ -1,61 +1,46 @@
...
 This plugin enables Azure Functions support within the Serverless Framework.
 
 [![Coverage Status](https://coveralls.io/repos/github/serverless/serverless-azure-functions/badge.svg)](https://coveralls.io/github/serverless/serverless-azure-functions)
 
-## Getting started
+## Quickstart
 
 ### Pre-requisites
 
 1. Node.js v6.5.0+ *(this is the runtime version supported by Azure Functions)*
```
mydiemho: should follow markdown list syntax and use 

```
1. one
1. two
1. three
```

README.md
```diff
@@ -65,58 +50,146 @@ To clean up files generated from the build, you can simply run:
...
+
+This will remove the `{functionName}.js` handler and remove the function from `serverless.yml`
+
+*Note: Add & remove currently only support HTTP triggered functions. For other triggers, you will need to update `serverless.yml` manually
+
+### Advanced Authentication
+
+The getting started walkthrough illustrates the interactive login experience, which is recommended when getting started. However, for more robust use, a [service principal](https://docs.microsoft.com/en-us/azure/active-directory/develop/app-objects-and-service-principals) is recommended for authentication.
+
+##### Creating a Service Principal
```
mydiemho: can you add reference to use the script I wrote?

tbarlow12: We should talk about where that should live. I don't know if it's in the quickstart, just because we expose that script in the `scripts` section of `package.json`, and it would be kinda awkward for people that were consuming the plugin to drill into `node_modules` to go run the script. If that script were available in the project template, then I think it would make sense to be here. I think that would go in other more detailed dev docs, just because we'd have to have other instructions like installing `jq`. Open to chat about where to put that in the future, but I'll probably leave it out for now

README.md
tbarlow12: The [markdown guide](https://www.markdownguide.org/basic-syntax/#ordered-lists) lists the actual numbered version as the first style, but they both render to the same thing. Also, the VS Code extension auto-generates the next number with a new line.

PR #205, opened 2019-07-17
feat: Deploy pre-existing package with `-p` CLI option

tbarlow12:
- [x] Allow `-p` or `--package` param to specify package to deploy
- [x] Log in before deployment rather than packaging
- [x] Don't run deployment if local package does not exist

Resolves [AB#500]


src/plugins/login/azureLoginPlugin.ts
```diff
@@ -14,6 +14,7 @@ export class AzureLoginPlugin extends AzureBasePlugin<AzureLoginOptions> {
 
     this.hooks = {
       "before:package:initialize": this.login.bind(this),
```
mydiemho: this line should be remove

src/services/functionAppService.test.ts
```diff
@@ -248,10 +283,19 @@ describe("Function App Service", () => {
     }, slsService["artifact"])
   });
 
-  it("throws an error with no zip file", async () => {
+  it("uses default name when no artifact provided", async () => {
```
mydiemho: previous tests was testing a different things, why is it being replace?

tbarlow12: This test was wrong. It was throwing an error, but for the wrong reason. Now that `getFunctionZipFile` is public, I can test it directly, which is why I renamed the test.

src/services/functionAppService.test.ts
```diff
@@ -222,6 +226,37 @@ describe("Function App Service", () => {
...
+    const expectedUploadUrl = `https://${scmDomain}/api/zipdeploy/`;
+
+    const sls = MockFactory.createTestServerless();
+    delete sls.service["artifact"]
+    const service = createService(sls);
+    await service.uploadFunctions(app);
+
+    const defaultArtifact = path.join(".serverless", `${sls.service.getServiceName()}.zip`);
+
+    expect((FunctionAppService.prototype as any).sendFile).toBeCalledWith({
```
mydiemho: nit: I would extract the `toBeCalledWith` arguments out just so it's easier to read

src/services/functionAppService.test.ts
```diff
@@ -222,6 +226,37 @@ describe("Function App Service", () => {
...
+      uri: expectedUploadUrl,
+      json: true,
+      headers: {
+        Authorization: `Bearer ${variables["azureCredentials"].tokenCache._entries[0].accessToken}`,
+        Accept: "*/*",
+        ContentType: "application/octet-stream",
+      }
+    }, defaultArtifact);
+    const expectedArtifactName = service.getDeploymentName().replace("rg-deployment", "artifact");
+    expect((AzureBlobStorageService.prototype as any).uploadFile).toBeCalledWith(
```
mydiemho: nit: extract `AzureBlobStorageService.prototype as any`

PR #207, opened 2019-07-18
feat: specify subscription ID in serverless yaml

njm3754:


src/plugins/login/azureLoginPlugin.ts
```diff
@@ -31,9 +31,8 @@ export class AzureLoginPlugin extends AzureBasePlugin<AzureLoginOptions> {
     try {
       const authResult = await AzureLoginService.login();
       this.serverless.variables["azureCredentials"] = authResult.credentials;
-      // Use environment variable for sub ID or use the first subscription in the list (service principal can
-      // have access to more than one subscription)
-      this.serverless.variables["subscriptionId"] = this.options.subscriptionId || process.env.azureSubId || authResult.subscriptions[0].id;
+      // Use environment variable for sub ID or use the first subscription in the list (service principal can have access to more than one subscription)
+      this.serverless.variables["subscriptionId"] = this.options.subscriptionId || process.env.azureSubId || this.serverless.service.provider["subscriptionId"] || authResult.subscriptions[0].id;
```
wbreza: I would like to see a new unit test to validate this change before you go ahead and merge.

njm3754: Added one test case. The test case when this.serverless.service.provider["subscriptionId"] is null has not been added since one of the previous test cases ("Uses the default subscription ID") already validates that behavior.

PR #208, opened 2019-07-19
feat: List deployments for rollback when no timestamp given

tbarlow12:
Rollback requires a timestamp. Rather than forcing users to run `sls deploy list` to generate the deployments they can choose from, `sls rollback` with no args also generates the same list.

Resolves [AB#493]


tbarlow12:
Rollback requires a timestamp. Rather than forcing users to run `sls deploy list` to generate the deployments they can choose from, `sls rollback` with no args also generates the same list.

Resolves [AB#493]

PR #211, opened 2019-07-19
fix: apim uses shortened region names

pjlittle:
When generating an APIM resource name, the service now
uses the shortened resource name, consistent with other resources.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "serverless-azure-functions",
-  "version": "1.0.0-6",
+  "version": "1.0.0-7",
   "description": "Provider plugin for the Serverless Framework v1.x which adds support for Azure Functions.",
   "license": "MIT",
   "main": "./lib/index.js",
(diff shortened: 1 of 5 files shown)


src/services/apimService.ts
```diff
@@ -28,7 +29,7 @@ export class ApimService extends BaseService {
     }
 
     if (!this.apimConfig.name) {
-      this.apimConfig.name = `${this.config.provider.prefix}-${this.config.provider.region}-${this.config.provider.stage}-apim`
+      this.apimConfig.name = `${this.config.provider.prefix}-${Utils.createShortAzureRegionName(this.config.provider.region)}-${this.config.provider.stage}-apim`
```
wbreza: We should call into Apim Resource.getResourceName instead. This is what is used when resource is initially generated.

pjlittle: Updated to use the exiting function.

pjlittle:
Bumped prerelease version in `package.json`

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "serverless-azure-functions",
-  "version": "1.0.0-6",
+  "version": "1.0.0-7",
```
mydiemho: also need to update `package-lock.json`

pjlittle: Good catch. fixed.

PR #212, opened 2019-07-23
feat: Path from x-azure-settings is included in function.json

tbarlow12:
In serverless.yml, you can now specify an API route within `x-azure-settings`, which will be added to the `function.json` file.

Resolves [AB#552]

### DIFF
--- a/src/services/packageService.test.ts
+++ b/src/services/packageService.test.ts
@@ -9,12 +9,13 @@ import { FunctionMetadata } from "../shared/utils";
 describe("Package Service", () => {
   let sls: Serverless;
   let packageService: PackageService;
+  const functionRoute = "myRoute";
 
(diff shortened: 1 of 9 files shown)


src/services/packageService.test.ts
```diff
@@ -71,17 +73,22 @@ describe("Package Service", () => {
...
         functionsJson: {},
       },
     };
 
     const expectedFolderPath = path.join(sls.config.servicePath, functionName);
     const expectedFilePath = path.join(expectedFolderPath, "function.json");
+    const expectedFunctionJson = {
+      entryPoint: "handler",
+      scriptFile: "src/handlers/hello",
+      route: "myRoute",
```
mydiemho: this should be in the event level

PR #214, opened 2019-07-24
feat: Spawn offline process with `sls offline` command

tbarlow12:
- [x] `sls offline` builds Azure Functions files and spawns `func host start` in the same shell
- [x] `sls offline start` spawns `func host start` in the same shell
- [x] `BaseService` has `spawn` function that spawns and waits for child process in shell (routes stdout and stderr to internal logger)
- [x] Using `mock-spawn` library to mock spawned child processes
- [x] Add invocation commands to config constants

Resolves [AB#569]


src/services/baseService.ts
```diff
@@ -193,6 +198,50 @@ export abstract class BaseService {
...
   }
 
+  /**
+   * Spawn a Node child process with predefined environment variables
+   * @param command CLI Command - NO ARGS
+   * @param args Array of arguments for CLI command
+   * @param env Additional environment variables to be set in addition to
+   * predefined variables in `serverless.yml`
+   */
+  protected spawn(command: string, args?: string[], env?: any): Promise<void> {
```
wbreza: I don't think every service would need this.  I'd suggest we just move it into the offline service?

tbarlow12: I guess that's true... This would never be used in a CI/CD process, only by local devs. So it makes sense that it would only live in offline

src/services/offlineService.ts
```diff
@@ -50,9 +52,70 @@ export class OfflineService extends BaseService {
...
+  /**
+   * Spawn a Node child process with predefined environment variables
+   * @param command CLI Command - NO ARGS
+   * @param args Array of arguments for CLI command
+   * @param env Additional environment variables to be set in addition to
+   * predefined variables in `serverless.yml`
+   */
+  private spawn(command: string, args?: string[], env?: any): Promise<void> {
+    env = {
+      ...this.serverless.service.provider["environment"],
```
pjlittle: :100:

pjlittle:
nice work, looks great :+1:

wbreza:
sls offline builds Azure Functions files and spawns func host start in the same shell
  sls offline start spawns func host start in the same shell
  BaseService has spawn function that spawns and waits for child process in shell (routes stdout and stderr to internal logger)
  Using mock-spawn library to mock spawned child processes
  Add invocation commands to config constants

src/services/offlineService.ts
```diff
@@ -50,9 +52,70 @@ export class OfflineService extends BaseService {
...
+      ...env
+    }
+    this.log(`Spawning process '${command} ${args.join(" ")}'`);
+    return new Promise((resolve, reject) => {
+      const childProcess = spawn(command, args, {env});
```
Gta-Cool: I don't know if it has been made on purpose (if yes, that requires maybe some documentation), but passing the `env` variable here as an option is causing the default value `process.env` to not be used (and more particularly the PATH in it). Taking the PATH from `process.env` if it hasn't been set as an additional environment variable could be a solution.

https://nodejs.org/docs/latest-v10.x/api/child_process.html#child_process_child_process_spawn_command_args_options

PR #215, opened 2019-07-25
fix: Add retry support when listing deployed functions

wbreza:
When listing the deployed functions on 1st time deployment it may take some additional time to initialize the function app and extract the functions.  This adds a retry policy that will retry the function listing multiple times until the function app is fully ready.

### DIFF
--- a/src/services/functionAppService.ts
+++ b/src/services/functionAppService.ts
@@ -106,7 +106,7 @@ export class FunctionAppService extends BaseService {
       return response.data.value.map((functionConfig) => functionConfig.properties);
     }
     catch (e) {
-      this.log("Unable to retrieve function app list");
+      this.log("-> Unable to retrieve function app list");
       return [];
     }
   }
--- a/src/services/functionAppService.ts
+++ b/src/services/functionAppService.ts
@@ -10,6 +10,7 @@ import { Guard } from "../shared/guard";
 import { ArmService } from "./armService";
 import { AzureBlobStorageService } from "./azureBlobStorageService";
 import { BaseService } from "./baseService";
+import { Utils } from "../shared/utils";
 
 export class FunctionAppService extends BaseService {
   private webClient: WebSiteManagementClient;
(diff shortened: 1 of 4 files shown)


src/services/functionAppService.ts
```diff
@@ -90,27 +91,53 @@ export class FunctionAppService extends BaseService {
...
+          throw new Error(listFunctionsResponse.data);
+        }
+
+        return listFunctionsResponse;
+      }, 10, 5000);
+
+      return response.data.value.map((functionConfig) => functionConfig.properties);
+    }
+    catch (e) {
+      this.log("-> Unable to retrieve function app list");
```
mydiemho: we should exit with error if we cannot retrieve the list of functions

tbarlow12: Throughout the plugin so far, we've just logged the error and exited gracefully

src/services/functionAppService.ts
```diff
@@ -90,27 +91,53 @@ export class FunctionAppService extends BaseService {
...
+      const response = await Utils.runWithRetry(async () => {
+        const getFunctionResponse = await this.sendApiRequest("GET", getFunctionUrl);
+
+        if (getFunctionResponse.status !== 200) {
+          this.log("-> Function app not ready. Retrying...")
+          throw new Error(response.data);
+        }
+
+        return getFunctionResponse;
+      }, 10, 5000);
```
pjlittle: maybe make 10 (retries?) and 5000 (ms) consts, since we reference them in two places?

pjlittle:
One small comment, otherwise LGTM! Hopefully this helps w/ some of the flakiness we've seen :shipit:

src/services/functionAppService.ts
wbreza: I agree with @mydiemho here.  If we are having a hard time getting the list of functions after deployment this can signal that there is some other slowdown which would likely cause next steps in the deployment process to fail.

src/services/functionAppService.ts
```diff
@@ -122,6 +151,23 @@ export class FunctionAppService extends BaseService {
     const uploadFunctionApp = this.uploadZippedArfifactToFunctionApp(functionApp, functionZipFile);
     const uploadBlobStorage = this.uploadZippedArtifactToBlobStorage(functionZipFile);
     await Promise.all([uploadFunctionApp, uploadBlobStorage]);
+
+
+    this.log("Deployed serverless functions:")
+    const serverlessFunctions = this.serverless.service.getAllFunctions();
+    const deployedFunctions = await this.listFunctions(functionApp);
+
+    // List functions that are part of the serverless yaml config
```
mydiemho: will there ever be a use case where there are deployed functions that are not in the yamL?

PR #218, opened 2019-07-29
fix: Inject current environment variables to spawn process

tbarlow12:
Not able to repro `ENOENT` error yet, but this PR should help with debugging.
- [x] Inject current environment variables to spawned process (most importantly, `PATH` so that an executable can be found)
- [x] Add additional log statements on errors for debugging purposes

See [this stack overflow answer](https://stackoverflow.com/a/27688805) for more details.

### DIFF
--- a/src/services/offlineService.ts
+++ b/src/services/offlineService.ts
@@ -1,4 +1,4 @@
-import { spawn } from "child_process";
+import { spawn, SpawnOptions } from "child_process";
 import fs from "fs";
 import Serverless from "serverless";
 import configConstants from "../config";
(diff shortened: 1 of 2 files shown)


src/services/offlineService.ts
```diff
@@ -62,59 +62,30 @@ export class OfflineService extends BaseService {
...
-    env = {
+  private spawn(command: string, spawnArgs?: string[]): Promise<void> {
+    if (process.platform === "win32") {
+      command += ".cmd";
+    }
+
+    const env = {
+      // Inherit environment from current process, most importantly, the PATH
+      ...process.env,
+      // Override any custom environment variables from serverless configuration
```
mydiemho: is this saying custom env variables will override yml or vice versa?

tbarlow12: The env variables from `serverless.yml` are king :)

PR #219, opened 2019-07-30
feat: Add Azure Naming Service

tbarlow12:
- [x] Added a centralized naming service for consistency and code re-use
- [x] Added `ArmResourceType` enum
- [x] Moved naming conventions out of base class and resource template files
- [x] Updated tests accordingly

So far, just moved naming tests to `azureNamingService.test.ts`, did not add any new tests. Will add more if needed. This is part of an effort to resolve [AB#597], regarding deployment names being too long. With this in place, it will be easier to maintain restrictions for naming conventions of each Azure Resource.


src/armTemplates/compositeArmTemplate.ts
```diff
@@ -36,13 +42,13 @@ export class CompositeArmTemplate implements ArmResourceTemplateGenerator {
     return template;
   }
 
-  public getParameters(config: ServerlessAzureConfig) {
+  public getParameters(config: ServerlessAzureConfig, namer: (resource: ArmResourceType) => string) {
```
wbreza: It doesn't feel right that the `getParameters(...)` function now requires the function to name the resource.  Since you have already added `ArmResourceType` enum - what if we have the ARM resource classes just take a dependency on the naming service?  And call the naming service as needed to generate the name?

src/models/armTemplates.ts
```diff
@@ -1,11 +1,23 @@
 import { ServerlessAzureConfig } from "./serverless";
 
+export enum ArmResourceType {
```
wbreza: Do we need to expose this?  Each ARM resource is already a class.  Can't we use the type of class as the identifier vs adding another property to get the ARM resource type?

src/services/azureNamingService.ts
```diff
@@ -0,0 +1,228 @@
...
+    return `${deploymentName
+      .replace("rg-deployment", "artifact")
+      .replace("deployment", "artifact")}.zip`;
+  }
+
+  /**
+   * Get name of Azure resource
+   * @param resource ARM Resource to name
+   */
+  public getResourceName(resource: ArmResourceType): string {
```
wbreza: I'm also uneasy about this change.  Currently the "naming convention" is controlled by the owning resource and the business logic / naming convention is close to the resource that solves other concerns about that resource.  Moving the resource specific implementation out of the resource into another service with an every growing number of functions seems off to me.  Let's discuss the overall direction here.

PR #221, opened 2019-07-31
feat: Added Azure Naming Service

tbarlow12:
- [x] Add `AzureNamingService`
- [x] Move naming utils to naming service
- [x] Added `getSafeResourceName` - used by `storageAccount` and `deploymentName`

Resolves [AB#597]


src/services/namingService.ts
```diff
@@ -0,0 +1,160 @@
...
+    const { prefix, region, stage } = config.provider
+    let name = [
+      prefix,
+      this.createShortAzureRegionName(region),
+      this.createShortStageName(stage),
+    ].join("-");
+    if (suffix) {
+      name += `-${suffix}`;
+    }
+    return name;
```
pjlittle: I can't think of a _great_ reason why, but would we want to `.toLowerCase()` this, just to normalize?  Also, guessing it's unsupported, but has anyone tried extended ascii or double-byte chars?

pjlittle:
LGTM - thanks for centralizing this logic and updating across the board :+1:

src/services/baseService.ts
```diff
@@ -43,7 +44,7 @@ export abstract class BaseService {
     this.deploymentConfig = this.getDeploymentConfig();
     this.deploymentName = this.getDeploymentName();
     this.storageAccountName = StorageAccountResource.getResourceName(
```
wbreza: Can likely just keep this as 1 line now.

PR #223, opened 2019-08-02
feat: Azure Pipelines CI pipeline & script

tbarlow12:
Simple CI check to start moving repo to Azure Devops. Will add release pipelines and scripts in future PR.

Resolves [AB#588]

### DIFF
--- a/pipelines/ci.yml
+++ b/pipelines/ci.yml
@@ -1,0 +1,19 @@
+# CI should only run when PRs are opened against dev/master.
+pr:
+  - master
+  - dev
+
+pool:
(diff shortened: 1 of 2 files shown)


pjlittle:
Looks good, so this will build, then we'll have a separate pipeline for publish? :shipit: 

@mydiemho - anything else to add here?

tbarlow12:
@pjlittle 
> Looks good, so this will build, then we'll have a separate pipeline for publish? :shipit:
> 
> @mydiemho - anything else to add here?

Exactly. Working on the publish pieces now. This just seemed like a low lift to verify that things are working

Issue #57, opened 2017-04-12
Incorrect Service Bus binding parsing


truztee:
This is the config I use in serverless.yaml
```yaml
functions:
  importJob:
     handler: handler.importJob
     events:
       - serviceBus:
         x-azure-settings:
           name: item
           queueName: import-queue
           accessRights: listen
           connection : ServiceBusConnection
```
This is the parsed binding result:

```javascript
{ type: 'serviceBusTrigger',
  direction: 'in',
  name: 'item',
  queueName: 'import-queue',
  topicName: 'mysbtopic',
  subscriptionName: 'mysubscription',
  connection: 'ServiceBusConnection',
  accessRights: 'listen' 
}
```

There are both queueName and topicName defined in the binding output along with subscriptionName which should not be there if queueName is defined.

pragnagopa:
This is a bug in parsing [bindings.json
](https://github.com/serverless/serverless-azure-functions/blob/master/shared/parseBindings.js). Parsing needs to consider rules for each of the bindings. For example:[ rules for ServiceBusTrigger](https://github.com/serverless/serverless-azure-functions/blob/master/shared/bindings.json#L967) indicate exclusive settings for queueName and topicName/SubscriptionName

truztee:
Would you consider using external library such as [validatorjs](https://github.com/skaterdav85/validatorjs)?

Also is there any milestone or any todo list for azure integration that we can follow? The documentation seems to be lack behind and others that would like to contribute cannot do so because we don't know which feature that still needs to be implemented.

arunshan:
@pragnagopa Why do we need to set the `topicName` and `subscriptionName` when using Service Bus Queues? Is it a requirement according to Azure docs? Thank You.

marcel-devdude:
Any news on this? Or any workaround? Currently it is not possible to create an service bus trigger based on queues instead of topics.

hanswang:
I wrote a custom plugin for this bug.
https://gist.github.com/hanswang/416c70d8786b1416ab01070cbbc54b8c.js

ChrisPanthera:
@hanswang how does one implement this plugin?

majezanu:
Hey, I have an issue related to this and https://github.com/serverless/serverless-azure-functions/issues/352, can you help me to fix it? There is a workaround?

PR #228, opened 2019-08-06
fix: issue fixed to generate the correct function.json for event hub and service bus queue and topic triggers

njm3754:


tbarlow12:
Change `fixed:` to `fix:` before committing. There should be some kind of unit test accompanying each change. It would probably be a test within the `packageService` or `bindings`.

src/shared/bindings.json
```diff
@@ -78,7 +78,7 @@
                     ]
                 },
                 {
-                    "name": "path",
+                    "name": "eventhubname",
```
wbreza: Is casing important here?  I thought they use `camelCasing` ? If its working that we are probaly good.

wbreza:
@njm3754 - Please make sure you associate the github issues with this fix.

src/shared/bindings.json
njm3754: It works with both I tested it again just to confirm. But I still used camel casing since it looks consistent and better and committed my new changes.

wbreza:
Resolves issue #57

PR #229, opened 2019-08-06
feat: Add code coverage to Azure DevOps

tbarlow12:
- [x] Removes code coverage from Travis and runs it within Azure DevOps pipeline
- [x] Only trigger pipeline on pull requests to dev or master

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,16 +32,13 @@ jobs:
       script:
         - npm run compile
         - npm run test:ci
-      after_success:
-        - npm run test:coverage
-        - cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js && rm -rf ./coverage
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -32,16 +32,13 @@ jobs:
       script:
```
mydiemho: can remove this completely since we'll be running the test on AzDO too

pipelines/ci.yml
```diff
@@ -17,3 +18,9 @@ steps:
   inputs:
     targetType: filePath
     filePath: ./scripts/build.sh
```
mydiemho: let's discuss offline about this build script, there's some duplicate work here

tbarlow12: @mydiemho sounds good. not super high pri right now anyway

PR #231, opened 2019-08-07
feat: Deploy function app code from blob storage URL

tbarlow12:
- [x] Allow configuration to specify whether or not to run from external package
- [x] Set function app setting `WEBSITE_RUN_FROM_PACKAGE` with SAS URL of package from blob storage
- [x] Get deployment configuration from `config.provider`
- [x] Upload to blob storage before deployment if `runFromBlobUrl` is specified

### DIFF
--- a/src/plugins/login/azureLoginPlugin.test.ts
+++ b/src/plugins/login/azureLoginPlugin.test.ts
@@ -52,6 +52,7 @@ describe("Login Plugin", () => {
   });
 
   it("calls login if azure credentials are not set", async () => {
+    unsetServicePrincipalEnvVariables();
     await invokeLoginHook();
     expect(AzureLoginService.interactiveLogin).toBeCalled();
     expect(AzureLoginService.servicePrincipalLogin).not.toBeCalled();
@@ -115,7 +116,7 @@ describe("Login Plugin", () => {
     expect(sls.variables["subscriptionId"]).toEqual("azureSubId");
     expect(sls.cli.log).toBeCalledWith("Using subscription ID: azureSubId");
   });
-  
+
   it("Uses the subscription ID specified in serverless yaml", async () => {
     const sls = MockFactory.createTestServerless();
     const opt = MockFactory.createTestServerlessOptions();
--- a/src/services/functionAppService.test.ts
+++ b/src/services/functionAppService.test.ts
@@ -17,7 +17,6 @@ jest.mock("./azureBlobStorageService");
 import { AzureBlobStorageService } from "./azureBlobStorageService"
 import configConstants from "../config";
 
-
 describe("Function App Service", () => {
   const app = MockFactory.createTestSite();
   const slsService = MockFactory.createTestService();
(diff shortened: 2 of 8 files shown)


src/services/functionAppService.ts
```diff
@@ -149,7 +149,13 @@ export class FunctionAppService extends BaseService {
 
     const functionZipFile = this.getFunctionZipFile();
     const uploadFunctionApp = this.uploadZippedArfifactToFunctionApp(functionApp, functionZipFile);
-    const uploadBlobStorage = this.uploadZippedArtifactToBlobStorage(functionZipFile);
+    // If `runFromBlobUrl` is configured, the artifact will have already been uploaded
+    const uploadBlobStorage = (this.deploymentConfig.runFromBlobUrl)
```
wbreza: To support rollback - don't we already ALWAYS upload the artifact to blob storage?

wbreza: Then its really just a configuration regarding whether the `RUN_FROM_PACKAGE` is set to 1 vs the SAS URL ???

src/services/functionAppService.ts
```diff
@@ -178,10 +184,18 @@ export class FunctionAppService extends BaseService {
...
-      ? await armService.createDeploymentFromConfig(this.config.provider.armTemplate)
-      : await armService.createDeploymentFromType(this.config.provider.type || "consumption");
-
+    const { armTemplate, type } = this.config.provider;
+    let deployment: ArmDeployment = armTemplate
+      ? await armService.createDeploymentFromConfig(armTemplate)
+      : await armService.createDeploymentFromType(type || "consumption");
+
+    if (this.deploymentConfig.runFromBlobUrl) {
+      await this.uploadZippedArtifactToBlobStorage(this.getFunctionZipFile());
```
wbreza: How can you upload the artifact to blob storage before the storage account is provisioned??

src/services/functionAppService.test.ts
```diff
@@ -37,6 +36,18 @@ describe("Function App Service", () => {
   const syncTriggersUrl = `${baseUrl}${app.id}/syncfunctiontriggers?api-version=2016-08-01`;
   const listFunctionsUrl = `${baseUrl}${app.id}/functions?api-version=2016-08-01`;
 
+  const baseConfigUrl = `${baseUrl}/subscriptions/azureSubId/resourceGroups/myResourceGroup/providers/Microsoft.Web/sites/Test/config/appsettings`
+  const apiVersion = "?api-version=2016-08-01"
+
+  const listConfigUrl = `${baseConfigUrl}/list${apiVersion}`;
+  const updateConfigUrl = baseConfigUrl + apiVersion;
+
```
wbreza: Nit: extra line

src/services/functionAppService.ts
```diff
@@ -226,6 +241,23 @@ export class FunctionAppService extends BaseService {
     return functionZipFile;
   }
 
+  public async updateFunctionAppSetting(functionApp: Site, setting: string, value: string) {
+    const baseUrl = `https://management.azure.com/subscriptions/${this.subscriptionId}` +
```
wbreza: Is there not a method in any of the imported SDK that we are using to use instead of manually crafting HTTP request?

src/services/functionAppService.ts
tbarlow12: Fixed.

src/plugins/login/azureLoginPlugin.test.ts
```diff
@@ -52,6 +52,7 @@ describe("Login Plugin", () => {
   });
 
   it("calls login if azure credentials are not set", async () => {
+    unsetServicePrincipalEnvVariables();
```
tbarlow12: This test was failing for some reason... Not sure how it got through the checks

PIC123: huh, that's weird, I thought it had been added in the tests, good thing they work now then

PR #232, opened 2019-08-08
fix: Removed service name from storage account name generator

wbreza:
When the storage account name is generated it includes a hash of the sls service name.  By default if you deploy multiple services/apps into the same resource group this causes multiple storage accounts to be created vs. reusing the same one.  By removing the service name from the storage account generation logic it simplifies the case to reuse the same account.  This value can still be overridden by specifying your own name in the sls yaml configuration.


src/services/namingService.ts
```diff
@@ -85,8 +84,8 @@ export class AzureNamingService {
       maxLength -= timestamp.length + suffix.length;;
 
       const name = (deploymentName) ? deploymentName.substr(0, maxLength)
-        : [ AzureNamingService.getSafeResourceName(config, maxLength), suffix ].join("-");
-      return [ name, timestamp ].join("-");
+        : [AzureNamingService.getSafeResourceName(config, maxLength), suffix].join("-");
```
mydiemho: this need to now include the service name otherwise this change will break deployment name

wbreza: Ah - didn't know `deployment name` was generated from this also.  Will address.

src/services/namingService.ts
```diff
@@ -42,37 +44,41 @@ export class AzureNamingService {
    * @param forbidden Regex for characters to remove from name. Defaults to non-alpha-numerics
    * @param replaceWith String to replace forbidden characters. Defaults to empty string
    */
-  public static getSafeResourceName(config: ServerlessAzureConfig, maxLength: number, resourceConfig?: ResourceConfig, forbidden = /\W+/g, replaceWith = "") {
+  public static getSafeResourceName(config: ServerlessAzureConfig, maxLength: number, resourceConfig?: ResourceConfig, suffix: string = "", forbidden: RegExp = /\W+/g, replaceWith: string = "") {
```
mydiemho: update jsdoc

src/services/namingService.ts
```diff
@@ -42,37 +44,41 @@ export class AzureNamingService {
...
       safeStage = safeStage.substr(0, partLength);
-      safeNameHash = safeNameHash.substr(0, partLength);
+      safeSuffix = safeSuffix.substr(0, partLength);
     }
 
-    return [safePrefix, safeRegion, safeStage, safeNameHash]
+    return [safePrefix, safeRegion, safeStage, safeSuffix]
       .join("")
-      .toLocaleLowerCase();
+      .toLowerCase();
```
mydiemho: why the switch?

src/services/namingService.test.ts
```diff
@@ -86,4 +86,60 @@ describe("Naming Service", () => {
     const actual = AzureNamingService.getNormalizedRegionName(expected);
     expect(actual).toEqual(expected);
   });
+
+  it("deployment name is generated correctly", () => {
```
mydiemho: I would put these 2 new tests in the deployment class test

src/services/namingService.test.ts
```diff
@@ -86,4 +86,60 @@ describe("Naming Service", () => {
...
+
+    const timestamp = Date.now().toString();
+
+    const deploymentName = AzureNamingService.getDeploymentName(config, `t${timestamp}`);
+    expect(deploymentName.length).toBeLessThanOrEqual(64);
+    expect(deploymentName).toContain(timestamp);
+    expect(deploymentName).toContain(AzureNamingService.createShortAzureRegionName(config.provider.region));
+    expect(deploymentName).toContain(createSafeString(config.provider.prefix));
+    expect(deploymentName).toContain(createSafeString(config.provider.stage));
+    expect(deploymentName).toContain(createSafeString(config.service));
```
mydiemho: also add a test for the right order <prefix-stage-region-service-timestamp>?

src/services/namingService.test.ts
```diff
@@ -86,4 +86,60 @@ describe("Naming Service", () => {
...
+        region: "westus",
+        stage: "dev",
+      },
+      service: "test-api"
+    };
+
+    const timestamp = Date.now().toString();
+
+    const deploymentName = AzureNamingService.getDeploymentName(config, `t${timestamp}`);
+    expect(deploymentName.length).toBeLessThanOrEqual(64);
```
mydiemho: replace all this with a call to `assertValidDeploymentName`

src/services/namingService.ts
wbreza: I don't believe it was ever intentional to use the locale specific version - it happens to be first option in autocomplete.  And since these names are not localizable and azure naming conventions expect standard english alpha numeric we don't need to use any local specific lowering.

src/services/namingService.test.ts
wbreza: Nice catch! - I had meant to do exactly that but didn't replace in all use cases.

src/services/namingService.test.ts
wbreza: They are related to the naming service - so going to keep these in here for now.

src/services/namingService.test.ts
wbreza: Added more explicit expectation for each test case to validate full expected name.

PR #233, opened 2019-08-08
feat: Handle SIGINT and cleanup offline files unless specified to skip

tbarlow12:
Cleans up offline files when user interrupts offline process. Allows user to skip cleanup with CLI option.

### DIFF
--- a/src/plugins/offline/azureOfflinePlugin.ts
+++ b/src/plugins/offline/azureOfflinePlugin.ts
@@ -30,7 +30,13 @@ export class AzureOfflinePlugin extends AzureBasePlugin {
             usage: "Start Azure Function app - assumes offline build has already occurred",
             lifecycleEvents: [
               "start"
-            ]
+            ],
+            options: {
+              nocleanup: {
+                usage: "Do not clean up offline files after finishing process",
+                shortcut: "n",
+              }
+            }
           },
           build: {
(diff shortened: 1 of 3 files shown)


src/services/offlineService.ts
```diff
@@ -76,10 +76,21 @@ export class OfflineService extends BaseService {
...
 
+      process.on("SIGINT", async () => {
+        if (this.getOption("nocleanup")) {
+          this.log("Skipping offline file cleanup...");
+          process.exit();
+        }
+        try {
+          await this.cleanup();
+        } catch(e) {}
+        process.exit();
```
wbreza: Consider moving the `process.exit()` into a `finally` block that can be reused for both condidtions.

```
try {
   // Do all the things
} finally {
  process.exit();
}
```

src/plugins/offline/azureOfflinePlugin.ts
```diff
@@ -30,7 +30,13 @@ export class AzureOfflinePlugin extends AzureBasePlugin {
             usage: "Start Azure Function app - assumes offline build has already occurred",
             lifecycleEvents: [
               "start"
-            ]
+            ],
```
wbreza: I'm assuming default behavior is to cleanup, correct?

tbarlow12: Yes

PR #237, opened 2019-08-09
Register missing dependency

medikoo:
[`src/test/mockFactory.js` directly relies on `@azure/ms-rest-js`](https://github.com/serverless/serverless-azure-functions/blob/962c619767870ce6a76cbf7a8cdacf15f7d0e658/src/test/mockFactory.ts#L4) while it's not listed in package.json deps, which implies a risk of package not being properly installed in some scenarios.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "xml": "^1.0.1"
   },
   "devDependencies": {
+    "@azure/ms-rest-js": "^1.8.7",
     "@babel/runtime": "^7.4.5",
     "@types/jest": "^24.0.13",
     "@types/jsonpath": "^0.2.0",


tbarlow12:
Looks great. If you can add the `fix: ` prefix to the commit message before squash/merging, we'd appreciate it :)

medikoo:
@tbarlow12 ok I've rebased it. Do you have any doc on CC rules (types to use etc.) you've settled on? (e.g. in one of the Serverless projects we use: https://docs.google.com/document/d/1hKUs3qt_aVp_PBI1UqvfaIqKma3jAJimEoGCRGGbOqs/ )

mydiemho:
> @tbarlow12 ok I've rebased it. Do you have any doc on CC rules (types to use etc.) you've settled on? (e.g. in one of the Serverless projects we use: https://docs.google.com/document/d/1hKUs3qt_aVp_PBI1UqvfaIqKma3jAJimEoGCRGGbOqs/ )

@medikoo that's good idea, will add a task to add a `CONTRIBUTE.md`

PR #241, opened 2019-08-13
perf: Skip upload to function app if configured to run from blob

tbarlow12:
Quick fix for not uploading zipped code to Function App if configured to only run from blob, and vice versa


pjlittle:
good catch.

PR #244, opened 2019-08-16
fix: Check for valid entries in token cache remove and find

tbarlow12:
Fails if user previously started interactive login process without finishing because it writes an invalid JSON file. This checks to make sure the parsed JSON is valid and that entries exists


pjlittle:
looks good :+1:

PR #245, opened 2019-08-16
feat: Add `hooks` type to base plugin, document plugin options for CLI help

tbarlow12:
- [x] Added `commands` to all plugins
- [x] Added `resourceGroup`, `stage`, `region` and `subscriptionId` to all CLI commands that need to log into Azure
- [x] Refactored `hooks` and `commands` to exist at the `BasePlugin` level
- [x] Added type `ServerlessHookMap`

Resolves AB#802 and AB#795

### DIFF
--- a/src/plugins/remove/azureRemovePlugin.ts
+++ b/src/plugins/remove/azureRemovePlugin.ts
@@ -3,10 +3,37 @@ import { ResourceService } from "../../services/resourceService";
 import { AzureBasePlugin } from "../azureBasePlugin";
 
 export class AzureRemovePlugin extends AzureBasePlugin {
-  public hooks: { [eventName: string]: Promise<any> };
 
   public constructor(serverless: Serverless, options: Serverless.Options) {
     super(serverless, options);
+
+    this.commands = {
+      remove: {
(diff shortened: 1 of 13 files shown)


src/plugins/remove/azureRemovePlugin.ts
```diff
@@ -3,10 +3,37 @@ import { ResourceService } from "../../services/resourceService";
...
 
 export class AzureRemovePlugin extends AzureBasePlugin {
-  public hooks: { [eventName: string]: Promise<any> };
 
   public constructor(serverless: Serverless, options: Serverless.Options) {
     super(serverless, options);
+
+    this.commands = {
+      remove: {
+        usage: "Remove service resource group (USE WITH CAUTION)",
```
mydiemho: eeep, is there a way to enforce confirmation with this command?

pjlittle:
LGTM

src/plugins/remove/azureRemovePlugin.ts
tbarlow12: Not currently, no

PR #251, opened 2019-08-19
fix: Remove VSCode project-level settings

medikoo:
I believe specific editor settings should not be committed into project.

In my case those settings collide with how I work with VS Code. I have some global settings, which I need to tweak in context of this project, still that's not possible without editing this file (as that in turn affects `git status`, which makes working with project repository cumbersome).

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,9 +1,0 @@
-{
-  "eslint.options": {
-    "configFile": "./.eslintrc.json"
-  },
-  "eslint.validate": [
-    "javascript",
-    "typescript"
-  ]
-}


medikoo:
> @medikoo - Shouldn't your global configuration settings live in a user profile configuration outside of the project's settings.json?

They are there, but I have a need to fine tune few of those settings for this project specifically, and that can't be done reliably with current setup.

e.g. in global settings I have set that I want to run formatting on save in case of all projects (In 99% of them it's desirable, and I don't want to apply these settings per project, that'll be cumbersome).

While in case of this project it introduces extra whitespace changes I don't want to have, therefore I want to keep it off here - and I can't do it without updating `.vscode/settings.json` (btw. I keep that file globally ignored in git, so it never ends committed to any repository)

I think the source of a problem here is, that ESLint (as of v6) via its `.eslintrc` doesn't support a way to state that all .ts files needs to be linted (it's supposed to be changed with v7 -> https://github.com/eslint/eslint/issues/10828), and I guess it's why you this is forced here via VS Code settings.

Still, I would rather let users configure editors on their own, as the way it's done here may conflict with developer personal preferences (as it's in my case).

PR #256, opened 2019-08-19
fix: Append 6-char resource group hash to storage account name

tbarlow12:
Because storage account names are global, the need to be unique across resource groups. This adds the first 6 characters of the `md5` resource group hash to the end of the storage account name.

Resolves AB#846


pjlittle:
looks good - given the constraints here, this change makes a lot of sense.

src/services/namingService.ts
```diff
@@ -77,9 +81,17 @@ export class AzureNamingService {
       safeSuffix = safeSuffix.substr(0, partLength);
     }
 
-    return [safePrefix, safeRegion, safeStage, safeSuffix]
+    const safeHash = md5(config.provider.resourceGroup).substr(0, hashLength);
+
+    const name = [safePrefix, safeRegion, safeStage, safeSuffix, safeHash]
```
mydiemho: nit: suffix should be at end

PR #285, opened 2019-08-29
feat: Azure Key Vault YAML setup support

PIC123:
It is bad practice to push code with secrets hard coded. Services such as keyvault allow users to securely store their keys and just reference them from their code. This change adds sets up a managed identity for the function app and sets permissions to allow the serverless app to get secrets.

AB#535


package.json
```diff
@@ -42,10 +42,12 @@
   "dependencies": {
     "@azure/arm-apimanagement": "^5.1.0",
     "@azure/arm-appservice": "^5.7.0",
+    "@azure/arm-keyvault": "^1.2.1",
     "@azure/arm-resources": "^1.0.1",
     "@azure/arm-storage": "^9.0.1",
     "@azure/ms-rest-nodeauth": "^1.0.1",
     "@azure/storage-blob": "^10.3.0",
+    "acorn": "^7.0.0",
```
wbreza: What is acorn used for?

src/plugins/identity/azureKeyVault.test.ts
```diff
@@ -0,0 +1,62 @@
+import Serverless from "serverless";
+import { MockFactory } from "../../test/mockFactory";
+import { invokeHook } from "../../test/utils";
+import { AzureApimServicePlugin } from "./azureApimServicePlugin";
+
+jest.mock("../../services/apimService");
+import { ApimService } from "../../services/apimService";
+
+describe("APIM Service Plugin", () => {
```
wbreza: Update references from `Apim` to `Keyvault`

src/plugins/identity/azureKeyVaultPlugin.ts
```diff
@@ -0,0 +1,28 @@
+import Serverless from "serverless";
+import { AzureBasePlugin } from "../azureBasePlugin";
+import { AzureKeyVaultService } from "../../services/azureKeyVaultService";
+
+export class AzureKeyVaultPlugin extends AzureBasePlugin {
+
```
wbreza: nit: Empty line

src/services/azureKeyVaultService.ts
```diff
@@ -0,0 +1,67 @@
+import Serverless from "serverless";
+import { BaseService } from "./baseService";
+import { FunctionAppService } from "./functionAppService";
+import { KeyVaultManagementClient } from "@azure/arm-keyvault";
+import { KeyPermissions, SecretPermissions } from "@azure/arm-keyvault/esm/models/index";
+
+
```
wbreza: nit: Empty line

src/services/azureKeyVaultService.ts
```diff
@@ -0,0 +1,67 @@
...
+    super(serverless, options);
+    this.funcApp = new FunctionAppService(serverless, options);
+  }
+
+  /**
+   * Sets the KeyVault policy for the function app to allow secrets permissions.
+   * @param keyVaultConfig Azure Key Vault settings
+   */
+  public async setPolicy(keyVaultConfig: AzureKeyVaultConfig) {
+    const subscriptionID = this.serverless.service.provider["subscriptionId"];
```
wbreza: Confirm this is a safe way to get the subscription id.  You can set this via the CLI flags as well.  Not sure if we normalized that all down into this config.

src/services/azureKeyVaultService.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    const func = await this.funcApp.get();
+    const identity = func.identity;
+
+    const keyVaultClient = new KeyVaultManagementClient(this.credentials, subscriptionID);
+    const vault = await keyVaultClient.vaults.get(keyVaultConfig.resourceGroup, keyVaultConfig.name);
```
wbreza: If the specified keyvault doesn't exist we should throw some sort of error.

src/services/azureKeyVaultService.ts
```diff
@@ -0,0 +1,67 @@
...
+        tenantId: identity.tenantId,
+        objectId: identity.principalId,
+        permissions: {
+          secrets: ["get" as SecretPermissions],
+        }
+      }
+      vault.properties.accessPolicies.push(newEntry);
+    } else {
+      policy.permissions = {
+        keys: ["get" as KeyPermissions],
```
wbreza: Thought we are removing `keys` from the permissions policy?

src/services/azureKeyVaultService.ts
```diff
@@ -0,0 +1,67 @@
...
+  public async setPolicy(keyVaultConfig: AzureKeyVaultConfig) {
+    const subscriptionID = this.serverless.service.provider["subscriptionId"];
+
+    const func = await this.funcApp.get();
+    const identity = func.identity;
+
+    const keyVaultClient = new KeyVaultManagementClient(this.credentials, subscriptionID);
+    const vault = await keyVaultClient.vaults.get(keyVaultConfig.resourceGroup, keyVaultConfig.name);
+    
+    const policy = vault.properties.accessPolicies.filter((val)=>{val.objectId === identity.principalId && val.tenantId === identity.tenantId})[0];
```
wbreza: You can use `find` instead of `filter`.  This will always return the first matched value or null.

src/services/azureKeyVaultService.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    if(!policy){
```
wbreza: Do we need the if/else condition?  What would happen if we just define the policy we want and call the `createOrUpdate` method below?  It should just get updated to the same settings if previously exists?

package.json
PIC123: it is a peer dependency of @azure/arm-keyvault

src/services/azureKeyVaultService.ts
PIC123: oh, good point, I'll have to look into this to see where we override it

package.json
wbreza: Really?  Weird!

src/plugins/identity/azureKeyVaultPlugin.test.ts
```diff
@@ -0,0 +1,53 @@
...
+    const plugin = new AzureKeyVaultPlugin(sls, options);
+
+    await invokeHook(plugin, "after:deploy:deploy");
+
+    expect(sls.cli.log).toBeCalledWith("Starting KeyVault service setup")
+    expect(setPolicy).toBeCalled();
+    expect(sls.cli.log).lastCalledWith("Finished KeyVault service setup")
+  });
+
+  it("does not call deploy API or deploy functions when \"keyVault\" not included in config", async () => {
```
wbreza: Is test name correct?  I would imagine it not calling `setPolicy` or similar, right?

src/services/azureKeyVaultService.test.ts
```diff
@@ -0,0 +1,112 @@
...
+      properties: {
+        accessPolicies: []
+      }
+    }
+  };
+
+  let serverless: Serverless = MockFactory.createTestServerless();
+
+  beforeEach(() => {
+    // KeyVaultManagementClient.prototype.vaults = {
```
wbreza: Removed commented code.

src/services/azureKeyVaultService.test.ts
```diff
@@ -0,0 +1,112 @@
...
+  it("is defined", () => {
+    expect(AzureKeyVaultService).toBeDefined();
+  });
+
+  it("can be instantiated", () => {
+    const service = new AzureKeyVaultService(serverless, options);
+    expect(service).not.toBeNull();
+  });
+
+  it("Throws an error if correct keyvault name is not specified", async () => {
```
wbreza: Test name seems off.  The test shows that the vault doesn't exist - not that it was not specified.

src/services/azureKeyVaultService.test.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    await expect(Vaults.prototype.createOrUpdate).toBeCalledWith(
+      keyVault.resourceGroup,
+      keyVault.name,
+      {
+        location: knownVaults["testVault"].location,
+        properties: knownVaults["testVault"].properties
+      }
+    );
+    // fail();
```
wbreza: Remove commented code.

src/services/azureKeyVaultService.ts
```diff
@@ -0,0 +1,57 @@
...
+   * @param keyVaultConfig Azure Key Vault settings
+   */
+  public async setPolicy(keyVaultConfig: AzureKeyVaultConfig) {
+    const subscriptionID = this.subscriptionId;
+
+    const func = await this.funcApp.get();
+    const identity = func.identity;
+
+    const keyVaultClient = new KeyVaultManagementClient(this.credentials, subscriptionID);
+    const vault = await keyVaultClient.vaults.get(keyVaultConfig.resourceGroup, keyVaultConfig.name).catch((e) => {throw new Error("Error: Specified vault not found")});
```
wbreza: Is the `.catch()` required here?  If you are using async/await it should throw an exception that you can catch with normal `catch` block.

src/armTemplates/resources/functionApp.ts
```diff
@@ -58,6 +58,9 @@ export class FunctionAppResource implements ArmResourceTemplateGenerator {
           "apiVersion": "2016-03-01",
           "name": "[parameters('functionAppName')]",
           "location": "[parameters('location')]",
+          "identity": {
```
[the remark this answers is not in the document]
PIC123: I am not sure, I don't know if there is any issue with always assigning the identity

turkoglusaffet:
How can we use it in a serverless.yml file? Is there any example for the topic?

PR #289, opened 2019-08-29
fix: Clean up downloaded artifact after rollback

### LINKED ISSUE #287: Rollback package not cleaned up when deploying directly to blob storage
Currently the mechanism for rolling back a function that deploys the code directly to blob storage (default behavior) is to download the blob to the `.serverless` folder and then redeploy the package. That should be cleaned up.

tbarlow12:
Remove the downloaded artifact from `.serverless` after deployed to service

Resolves #287

### DIFF
--- a/src/services/rollbackService.ts
+++ b/src/services/rollbackService.ts
@@ -8,6 +8,7 @@ import { BaseService } from "./baseService";
 import { FunctionAppService } from "./functionAppService";
 import { ResourceService } from "./resourceService";
 import { ArmDeployment } from "../models/armTemplates";
+import fs from "fs";
 
 /**
  * Services for the Rollback Plugin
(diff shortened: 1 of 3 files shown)


src/services/rollbackService.ts
```diff
@@ -98,6 +99,9 @@ export class RollbackService extends BaseService {
     const functionAppService = new FunctionAppService(this.serverless, this.options);
     const functionApp = await functionAppService.get();
     await functionAppService.uploadZippedArfifactToFunctionApp(functionApp, artifactPath);
+    if (fs.existsSync(artifactPath)) {
```
wbreza: Aren't we already cleaning up this folder before deployment?  Seems like we should maybe keep this around post deployment so developers can open up the zip more easily to inspect its contents??

wbreza: Nevermind - this is after a rollback - not deployment.

mydiemho: still think it's good to keep around even for rollback.  In case a rollback does not behave as expected, user can verify the deployed content

PR #290, opened 2019-08-29
fix: Combine rollback log statement into one string

tbarlow12:
Strings were being logged out of order at random for some reason. Seemed cleaner anyway to just combine them into one string and adjusted tests.

### DIFF
--- a/src/services/rollbackService.test.ts
+++ b/src/services/rollbackService.test.ts
@@ -76,11 +76,8 @@ describe("Rollback Service", () => {
     const options = {} as any;
     const service = createService(sls, options);
     await service.rollback();
-    expect(sls.cli.log).lastCalledWith(
-      deploymentString,
-      undefined,
-      undefined
-    );
+    const logCalls: any[][] = (sls.cli.log as any).mock.calls;
+    expect(logCalls[logCalls.length - 1][0].endsWith(deploymentString)).toBe(true);
   });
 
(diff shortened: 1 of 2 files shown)


mydiemho:
> Strings were being logged out of order at random for some reason. Seemed cleaner anyway to just combine them into one string and adjusted tests.

we might want to dig deeper as to why log is out of order, I am seeing this in other places too (see 4th line)

```
Serverless: Finished uploading blob
Serverless: -> Function App not ready. Retrying...
Serverless: -> Function App not ready. Retrying...
Serverless: -> Function package uploaded successfully
Serverless: -> Function App not ready. Retrying...
```

tbarlow12:
> > Strings were being logged out of order at random for some reason. Seemed cleaner anyway to just combine them into one string and adjusted tests.
> 
> we might want to dig deeper as to why log is out of order, I am seeing this in other places too (see 4th line)
> 
> ```
> Serverless: Finished uploading blob
> Serverless: -> Function App not ready. Retrying...
> Serverless: -> Function App not ready. Retrying...
> Serverless: -> Function package uploaded successfully
> Serverless: -> Function App not ready. Retrying...
> ```

That one I can see happening since it's just on a timer. Other functions might finish before. I think in particular, there might be a `Promise.all` in that section which would allow the two processes to run at the same time

Issue #292, opened 2019-08-30
Wrong resource name for appServicePlan and functionApp


stelone:
When the "name" of appServicePlan and functionApp is filled in the template, the resource created or updated have the wrong name. 

The serverless template look like this : 
```
service: dtp-dev-project-api
provider:
  name: azure
  runtime: nodejs8.11.1
  region: francecentral
  resourceGroup: pet-store-rg
  functionApp: 
    name: pet-store-api
  appServicePlan: 
    name: pet-store-plan
  appInsights: 
    name: pet-store-insights
  storageAccount: 
    name: pet-store-sa
```

The resource take appInsights name. 

I identify in the folder armTemplates the cause : 

AppServicePlanResource.ts : 
```
export class AppServicePlanResource implements ArmResourceTemplateGenerator {
    public static getResourceName(config: ServerlessAzureConfig) {
      return AzureNamingService.getResourceName(config, config.provider.**appInsights**, "asp");
    }
```

FunctionAppResource.ts : 
```
export class FunctionAppResource implements ArmResourceTemplateGenerator {
  public static getResourceName(config: ServerlessAzureConfig) {
    const safeServiceName = config.service.replace(/\s/g, "-");
    return AzureNamingService.getResourceName(config, config.provider.**appInsights**, safeServiceName);
  }
```

tbarlow12:
@stelone thank you for the issue & PR. The new beta should be out shortly.

PR #293, opened 2019-08-30
fix appServicePlan name and functionApp name

stelone:
When the "name" of appServicePlan and functionApp is filled in the template, the resource created or updated have the wrong name.

see #292


tbarlow12:
Closed and re-opened after a change in the AzDO pipeline

Issue #250, opened 2019-08-19
TypeError when on free-trial/pay-as-you-go subscription


medikoo:
When trying to deploy service when being logged in to _Free Trial_ or _Pay-As-You-Go_ subscription, the deployment crahes with:

```
Type Error ---------------------------------------------
 
  TypeError: Cannot read property 'id' of undefined
    at AzureLoginPlugin.login (/Users/medikoo/npm-packages/serverless-azure-functions/lib/plugins/login/azureLoginPlugin.js:37:192)

```

As we learned through testing, the workaround for that is to set manually `azureSubId`, `azureServicePrincipalTenantId`, `azureServicePrincipalClientId` and `azureServicePrincipalPassword` env variables.

Still it'll be good that plugin either handles that transparently, or exposes clear instructions to user (with what values what env vars needs to be assigned), instead of throwing non informative `TypeError`

medikoo:
> @medikoo are you still having issues here?

@tbarlow12 I don't think anything changed, right? I still need to do manual setup of env vars to get things working? Otherwise I didn't approach any new issues

medikoo:
Also error message doesn't give any lead.

At least for meantime it'll be good to change it so it's meaningful and provides instructions on how to fix the problem manually

tbarlow12:
@medikoo we're working with some people on the Azure product team to identify the issues here. We'll provide updates as we receive them

PR #294, opened 2019-08-30
refactor: Move subscriptionId logic to BaseService

tbarlow12:
- [x] Centralizes logic for getting subscription ID, similar to region, stage, prefix and resource group name
- [x] Adds error catching for empty subscriptions array from authentication (seen in #250)

### DIFF
--- a/src/models/serverless.ts
+++ b/src/models/serverless.ts
@@ -36,6 +36,7 @@ export interface ServerlessAzureProvider {
   region: string;
   stage: string;
   name: string;
+  subscriptionId?: string;
   environment?: {
     [key: string]: any;
   };
--- a/src/services/baseService.test.ts
+++ b/src/services/baseService.test.ts
@@ -39,8 +39,11 @@ class MockService extends BaseService {
 
 describe("Base Service", () => {
   let service: MockService;
-  let sls: Serverless;
+  let serverless: Serverless;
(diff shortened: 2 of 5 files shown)


src/models/serverless.ts
```diff
@@ -36,6 +36,7 @@ export interface ServerlessAzureProvider {
   region: string;
   stage: string;
   name: string;
+  subscriptionId?: string;
```
wbreza: surprised we didn't have this in here already!

src/services/baseService.test.ts
```diff
@@ -202,36 +204,62 @@ describe("Base Service", () => {
...
+
+  it("sets subscription ID from CLI", async () => {
+    serverless.service.provider["subscriptionId"] = configSubscriptionId;
+    serverless.variables["subscriptionId"] = loginResultSubscriptionId
+    service = new MockService(serverless, { subscriptionId: cliSubscriptionId } as any);
+    expect(service.getSubscriptionId()).toEqual(cliSubscriptionId);
+    expect(serverless.service.provider["subscriptionId"]).toEqual(cliSubscriptionId);
+  });
+
+  it("sets subscription ID from CLI", async () => {
```
wbreza: Consider adding 1 more for the ENV  variable set as sub id

PR #301, opened 2019-09-03
fix: Move service initialization outside of plugins

tbarlow12:
Fixes error we found in testing serverless optional variables (`${opt:stage, 'dev'}`), which allow you to specify a param at deploy time, as well as a default value. Also fixes potentially more serious problems with setting names before the configuration is initialized.

### DIFF
--- a/src/plugins/offline/azureOfflinePlugin.ts
+++ b/src/plugins/offline/azureOfflinePlugin.ts
@@ -3,11 +3,9 @@ import { OfflineService } from "../../services/offlineService";
 import { AzureBasePlugin } from "../azureBasePlugin";
 
 export class AzureOfflinePlugin extends AzureBasePlugin {
-  private offlineService: OfflineService;
 
   public constructor(serverless: Serverless, options: Serverless.Options) {
     super(serverless, options);
-    this.offlineService = new OfflineService(this.serverless, this.options);
 
     this.hooks = {
       "before:offline:offline": this.azureOfflineBuild.bind(this),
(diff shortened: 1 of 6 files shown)


mydiemho:
> Fixes error we found in testing serverless optional variables (`${opt:stage, 'dev'}`), which allow you to specify a param at deploy time, as well as a default value. Also fixes potentially more serious problems with setting names before the configuration is initialized.

1. what is the error you see when testing ` (`${opt:stage, 'dev'}`),`?
1. What are the serious problems with setting names before configuration is initialized?

src/services/baseService.ts
```diff
@@ -256,8 +256,13 @@ export abstract class BaseService {
    * Overwrite values for resourceGroup, prefix, region and stage
    * in config if passed through CLI
    */
-  private setConfigFromCli() {
+  private setupConfig() {
     const { prefix, region, stage, subscriptionId } = this.config.provider;
+
+    if (stage.startsWith("${")) {
```
mydiemho: why is this needed?

src/plugins/offline/azureOfflinePlugin.ts
```diff
@@ -60,14 +58,17 @@ export class AzureOfflinePlugin extends AzureBasePlugin {
   }
 
   private async azureOfflineBuild(){
-    await this.offlineService.build();
+    const offlineService = new OfflineService(this.serverless, this.options);
```
mydiemho: probably not too much of a difference, but what are we given up creating a new service everytime?

src/services/baseService.ts
tbarlow12: Leftover from before. Good callout :)

src/plugins/offline/azureOfflinePlugin.ts
tbarlow12: We can't initialize a `BaseService` within the constructor of a plugin. The constructor of `BaseService` assumes that the core framework has made whatever config substitutions it needs to do

wbreza: This looks good enough for now - but I feel the `BaseService` is starting to do too much and we should consider scaling it back at some point.  Ex) It does a lot of config initialization and if you think about all the different services we instantiate in the course of calling a plugin the same code is getting run over and over.

Issue #300, opened 2019-09-03
More specific error message for failed deployment


tbarlow12:
When a deployment fails, we get the "at least one resource deployment failed" message. It would be nice to just pull the error message from the deployment and display that instead.

martino:
I found the reason in the events section of the resource groups in the azure dashboard

PR #306, opened 2019-09-04
feat: More detailed error message upon ARM deployment failure

tbarlow12:
Logs a more detailed error message when an ARM template deployment fails:

image

### DIFF
--- a/src/models/azureProvider.ts
+++ b/src/models/azureProvider.ts
@@ -18,3 +18,9 @@ export interface FunctionMetadata {
   handler: string;
   events: FunctionEvent[];
 }
+
+export interface DeploymentExtendedError {
+  code: string;
+  message: string;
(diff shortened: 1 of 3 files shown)


src/models/azureProvider.ts
```diff
@@ -18,3 +18,9 @@ export interface FunctionMetadata {
   handler: string;
   events: FunctionEvent[];
 }
+
+export interface DeploymentExtendedError {
```
mydiemho: Why `DeploymentExtendedErorr`?  are we using `DeploymentError` already elsewhere?

tbarlow12: This is just an augmentation of the `DeploymentExtended` type (not defined by us). It has an `error` property that is not specified in the type.

tbarlow12:
Closes #300

PR #308, opened 2019-09-05
fix: Updates default http output binding

wbreza:
This updates the default value for the name of the HTTP output binding from `$return` to `$res` to work around an issue that existing within the Azure core tools http runtime..

A bug exists in the Azure core tools http runtime that evaluates `$return` binding incorrectly which causes serialization issues when returning arrays within an HTTP response.
https://github.com/Azure/azure-functions-nodejs-worker/issues/228


src/shared/bindings.ts
```diff
@@ -57,15 +57,12 @@ export class BindingUtils {
...
+  public static getHttpOutBinding() {
     const binding = {};
 
     binding[constants.type] = "http";
     binding[constants.direction] = constants.outDirection;
-    binding[constants.name] = "$return";
-    if (bindingUserSettings[constants.webHookType]) {
-      binding[constants.name] = "res";
-    }
+    binding[constants.name] = "res";
```
PIC123: Do we always want it to be res, or just default to it and still have the option of choosing the $return?

jon-lewis: I need the option of choosing $return instead of res.

wbreza:
@jon-lewis Even though the default has changed to `res` this can always be overriden in your serverless.yml as needed for your implementation.

You can modify your handler definition to include the output binding.  When you handler specifies its own output binding then the default is ignored.
```
  health:
    handler: src/handlers/health.handler
    events:
      - http: true
        x-azure-settings:
          methods:
            - GET
          authLevel : function
      - http: true
        x-azure-settings:
          direction: out
          name: res
```

Hope this helps.

jon-lewis:
Perfect. I did test this out and the following settings worked for me.

`
- http: true
  x-azure-settings:
    direction: out
    name: $return
`

Issue #299, opened 2019-09-03
Global naming of APIM


tbarlow12:
Similar to storage account name, the deployment of API management fails if there is another APIM with the same prefix, stage and region within the subscription

PR #310, opened 2019-09-05
fix: add resource group name hash to resource names

PIC123:
fix: add resource group name hash to resource names

Azure has naming requirements for resources, and for some, requires the names to be unique across azure. Currently on deployments, the plugin doesn't add a more unique hash to the name, so it is possible collisions can occur. By adding the resource group name hash to the resource name, we can avoid this.

AB#1002

### DIFF
--- a/src/services/namingService.ts
+++ b/src/services/namingService.ts
@@ -26,7 +26,7 @@ export class AzureNamingService {
       this.createShortStageName(stage),
     ].join("-");
 
-    if(includeHash) {
+    if(includeHash && config.provider.resourceGroup) {
       name += `-${md5(config.provider.resourceGroup).substr(0, configConstants.resourceGroupHashLength)}`
     }
 
--- a/src/services/namingService.ts
+++ b/src/services/namingService.ts
@@ -3,6 +3,13 @@ import { Guard } from "../shared/guard"
 import configConstants from "../config";
 import md5 from "md5";
 
+export interface AzureNamingServiceOptions {
+  config: ServerlessAzureConfig;
+  resourceConfig?: ResourceConfig;
+  suffix?: string;
+  includeHash?: boolean;
+}
+
 export class AzureNamingService {
 
   /**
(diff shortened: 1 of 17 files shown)


src/services/namingService.ts
```diff
@@ -26,6 +26,10 @@ export class AzureNamingService {
       this.createShortStageName(stage),
     ].join("-");
 
+    if(includeHash && config.provider.resourceGroup) {
```
tbarlow12: I think if `config.provider.resourceGroup` is not defined, there are bigger problems here. It would probably be good to throw an error as opposed to just skipping the addition of the hash

src/services/namingService.ts
```diff
@@ -14,7 +14,7 @@ export class AzureNamingService {
    * @param resourceConfig The serverless resource configuration
    * @param suffix Optional suffix to append on the end of the generated name
    */
-  public static getResourceName(config: ServerlessAzureConfig, resourceConfig?: ResourceConfig, suffix?: string) {
+  public static getResourceName(config: ServerlessAzureConfig, resourceConfig?: ResourceConfig, suffix?: string, includeHash: boolean = true) {
```
wbreza: These flags and extra params are getting out of hand.  Let's convert this into a typed  `options` object

src/services/namingService.ts
```diff
@@ -46,11 +61,11 @@ export class AzureNamingService {
    * @param forbidden Regex for characters to remove from name. Defaults to non-alpha-numerics
    * @param replaceWith String to replace forbidden characters. Defaults to empty string
    */
-  public static getSafeResourceName(config: ServerlessAzureConfig, maxLength: number, resourceConfig?: ResourceConfig, suffix: string = "", includeHash = false) {
+  public static getSafeResourceName(options: AzureNamingServiceOptions, maxLength: number) {
```
wbreza: Any reason `maxLength` isn't in the options?

PIC123: Only since the regular get resource name func didn't need it, but it makes more sense to add

wbreza:
Resolves #299

PR #312, opened 2019-09-05
feat: APIM feature updates

wbreza:
APIM configuration can easily bloat your serverless configuration yaml.  This PR reduces a lot of the boilerplate APIM configuration in many use cases except for when direct customization is required.

- [x] Infers APIM operation configuration from HTTP binding
- [x] Will automatically generate API just by setting `apm: true` in serverless yaml
- [x] Will automatically generate default backend pointing to the `/api` endpoint of your function app
- [x] Now supports configuring multiple API's in a single configuration
- [x] Now supports configuring multiple Backends in a single configuration
- [x] Ability to link an APIM operation to a specific API and backend


src/plugins/apim/azureApimPlugin.ts
```diff
@@ -2,12 +2,26 @@ import Serverless from "serverless";
...
     this.hooks = {
-      "after:deploy:deploy": this.deploy.bind(this)
+      "after:deploy:deploy": this.deploy.bind(this),
+      "deploy:apim:apim": this.deploy.bind(this),
+    };
+
+    this.commands = {
+      deploy: {
+        usage: "Deploys APIM",
+        lifecycleEvents: ["deploy"],
```
tbarlow12: I think this is what's causing the deployment to run twice.

tbarlow12: If we want `sls deploy apim` to be an available command, I think we'd need to add it as a sub-command to the `AzureDeployPlugin` rather than the APIM plugin

wbreza: I'll remove this and try again - i mostly put it in to aid testing

src/services/baseService.ts
```diff
@@ -196,7 +194,7 @@ export abstract class BaseService {
    * Log message to Serverless CLI
    * @param message Message to log
    */
-  protected log(message: string, options?: ServerlessLogOptions, entity?: string,) {
+  protected log(message: string, options?: ServerlessLogOptions, entity?: string, ) {
```
tbarlow12: nit: Remove last comma and empty space from method header

src/services/loginService.ts
```diff
@@ -40,22 +40,21 @@ export class AzureLoginService extends BaseService {
...
       authResp.subscriptions = fileTokenCache.listSubscriptions();
     }
 
     return authResp;
-
   }
 
   public async servicePrincipalLogin(clientId: string, secret: string, tenantId: string, options: AzureTokenCredentialsOptions): Promise<AuthResponse> {
-    return loginWithServicePrincipalSecretWithAuthResponse(clientId, secret, tenantId, options);
+    return await loginWithServicePrincipalSecretWithAuthResponse(clientId, secret, tenantId, options);
```
tbarlow12: Good catch

PR #316, opened 2019-09-06
refactor: Update service principal env variable names

tbarlow12:
Updates Azure Service Principal naming convention.

Will need to update Azure Pipelines before merging this change.


wbreza:
@tbarlow12 We'll also need to make sure we update the docs in serverless core repo.

PR #318, opened 2019-09-06
fix: Create default .azure directory if not present

### LINKED ISSUE #317: Error when performing interactive login without a .azure directory
If a user performs an interactive login without having a .azure directory in their home directory, the deployment fails and nothing can be deployed. Would prefer that the plugin addressed this without user intervention.

PIC123:
fix: Create default .azure directory if not present

The pluging currently saves credentials from interactive login in the .azure directory by default. If the user doesn't have a .azure directory, current;y the plugin hits an error. This fix checks to make sure the .azure directory exists and creates it if not. 

Closes #317


src/plugins/login/utils/simpleFileTokenCache.ts
```diff
@@ -11,6 +11,9 @@ export class SimpleFileTokenCache implements adal.TokenCache {
   private subscriptions: any[] = [];
 
   public constructor(private tokenPath: string = DEFAULT_SLS_TOKEN_FILE) {
+    if(tokenPath === DEFAULT_SLS_TOKEN_FILE && !fs.existsSync(CONFIG_DIRECTORY)) {
```
wbreza: nit: spacing - Run auto-format

PR #324, opened 2019-09-09
fix: Throw original error if no previous deployments

tbarlow12:
If a first-time deployment fails, it will crash with a weird error because the last deployment is undefined. This handles that scenario.


src/services/armService.ts
```diff
@@ -125,10 +125,13 @@ export class ArmService extends BaseService {
       return result;
     } catch (err) {
       const lastDeployment = await resourceService.getLastDeployment();
```
mydiemho: I think `previousDeployment` is a better term, cause `last` could have multiple meaning

PR #325, opened 2019-09-09
fix: Exclude ARM params with null values

tbarlow12:
- [x] Eliminate explicit `type` from valued paramaters
- [x] Remove any parameters whose value is `null` before deployment
- [x] Make `type` optional on `ArmParameters`

### DIFF
--- a/src/services/armService.ts
+++ b/src/services/armService.ts
@@ -102,6 +102,12 @@ export class ArmService extends BaseService {
       return;
     }
 
+    for (const key of Object.keys(deployment.parameters)) {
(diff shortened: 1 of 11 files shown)


src/services/armService.ts
```diff
@@ -102,6 +102,12 @@ export class ArmService extends BaseService {
       return;
     }
 
+    for (const key of Object.keys(deployment.parameters)) {
+      if (!deployment.parameters[key].value) {
+        deployment.parameters[key] = undefined;
```
wbreza: We can delete the param using the following: `delete deployment.parameters[key];`

PR #339, opened 2019-09-17
fix: Sync triggers on external package deployment

tbarlow12:
## What did you implement:

Fix updating of function app settings (SDK call wasn't working) and syncing triggers for function apps running from external package.

## How did you implement it:

- Updated the `syncTriggers` function to use [the correct URL](https://docs.microsoft.com/en-us/azure/azure-functions/functions-deployment-technologies#trigger-syncing)
- Updated the `updateFunctionAppSetting` function to just make a rest API call instead of using the SDK. The SDK call now gives the error message: `Error: Failed to update function app settings: Error: {"Code":"BadRequest","Message":"SiteAppSettings object is not present in the request body...`. I checked with the documentation and found matching GitHub issues. Rather than the SDK call, just making a REST request according to [the docs](https://docs.microsoft.com/en-us/rest/api/appservice/webapps/updateapplicationsettings)

## How can we verify it:

Deploy a function app with the following configuration:

```yaml
# serverless.yml
provider:
    deployment:
        external: true
```

## Todos:

       **Validate via `npm test`**
- [x] Write documentation
- [x] Make sure code coverage hasn't dropped

### DIFF
--- a/src/services/functionAppService.ts
+++ b/src/services/functionAppService.ts
@@ -63,13 +63,20 @@ export class FunctionAppService extends BaseService {
     return await this.sendApiRequest("DELETE", deleteFunctionUrl);
   }
 
-  public async syncTriggers(functionApp: Site) {
+  public async syncTriggers(functionApp: Site, properties: { [propertyName: string]: string }) {
     Guard.null(functionApp);
 
     this.log("Syncing function triggers");
 
-    const syncTriggersUrl = `${this.baseUrl}${functionApp.id}/syncfunctiontriggers?api-version=2016-08-01`;
-    return await this.sendApiRequest("POST", syncTriggersUrl);
+    const syncTriggersUrl = `${this.baseUrl}/subscriptions/${this.subscriptionId}` +
+      `/resourceGroups/${this.resourceGroup}/providers/Microsoft.Web/sites/${functionApp.name}` +
+      "/syncfunctiontriggers?api-version=2016-08-01";
+
+    try {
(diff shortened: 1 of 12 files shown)


tbarlow12:
> Since you are modifying a yaml config property this should be treated as a breaking change unless we support both `runFromBlobUrl` as well as `external`.

@wbreza I can add a backfill similar to the `location` and `region`. Good catch

src/services/functionAppService.ts
```diff
@@ -257,7 +268,16 @@ export class FunctionAppService extends BaseService {
   public async updateFunctionAppSetting(functionApp: Site, setting: string, value: string) {
     const { properties } = await this.webClient.webApps.listApplicationSettings(this.resourceGroup, functionApp.name);
     properties[setting] = value;
-    await this.webClient.webApps.updateApplicationSettings(this.resourceGroup, functionApp.name, properties);
```
tbarlow12: This SDK was failing even though the params were correct. Would like to use an SDK call instead of making our own rest API request, but needed to do so in order to unblock

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "serverless-azure-functions",
-  "version": "1.0.0",
+  "version": "1.0.1-1",
```
wbreza: You can likely revert this since master has now been merged back into dev.

src/services/functionAppService.ts
wbreza: Do we know why it was failing or what is the difference between the SDK call vs. the manual REST API call?

PR #354, opened 2019-09-30
fix: Sort deployments in descending order and fix APIM arm template

tbarlow12:
## What did you implement:

- Updated parameter name in APIM arm template
- Fixed bug of sorting deployments in ascending order, when it should have been descending. This would have pretty serious consequences, because it means that the comparison of ARM templates would always be targeting the first ever deployment, not the most recent.
- Because the `sort()` function sorts the array in place, this bug was not being detected by the tests. Updated `resourceService` tests to create copies of the array rather than using the original reference when testing the validity of the result.

## How did you implement it:

- Flipped the timestamp comparison operator
- Updated the string in `apim.ts`

## How can we verify it:

- Deploy a service
- Make changes to an environment variable
- Deploy it again (should deploy ARM template)
- Change back to the original configuration
- Deploy it again (should deploy ARM template)

Without this fix, the last step would not have deployed the ARM template since it would be identical to the first deployment ever.

## Todos:

       **Validate via `npm test`**
- [x] Write documentation
- [x] Make sure code coverage hasn't dropped

### DIFF
--- a/src/armTemplates/resources/apim.ts
+++ b/src/armTemplates/resources/apim.ts
@@ -61,7 +61,7 @@ export class ApimResource implements ArmResourceTemplateGenerator {
           "location": "[parameters('location')]",
           "sku": {
             "name": "[parameters('apimSkuName')]",
-            "capacity": "[parameters('apimCapacity')]"
+            "capacity": "[parameters('apimSkuCapacity')]"
           },
           "properties": {
             "publisherEmail": "[parameters('apimPublisherEmail')]",
--- a/src/services/resourceService.test.ts
+++ b/src/services/resourceService.test.ts
@@ -7,11 +7,14 @@ import { ResourceService } from "./resourceService";
 jest.mock("@azure/arm-resources")
 
 describe("Resource Service", () => {
-  let deployments: DeploymentsListByResourceGroupResponse;
+  
+  /**
+   * List of deployments in ASCENDING order (oldest first)
+   */
+  const deployments: DeploymentsListByResourceGroupResponse = MockFactory.createTestDeployments(5, true);
   const template = "myTemplate";
 
   beforeEach(() => {
-    deployments = MockFactory.createTestDeployments(5, true);
     ResourceManagementClient.prototype.resourceGroups = {
       createOrUpdate: jest.fn(),
       deleteMethod: jest.fn(),
@@ -19,7 +22,7 @@ describe("Resource Service", () => {
 
     ResourceManagementClient.prototype.deployments = {
       deleteMethod: jest.fn(),
-      listByResourceGroup: jest.fn(() => Promise.resolve(deployments)),
+      listByResourceGroup: jest.fn(() => Promise.resolve([...deployments])),
       exportTemplate: jest.fn(() => Promise.resolve(template)),
     } as any;
   });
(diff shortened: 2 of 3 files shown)


src/services/resourceService.ts
```diff
@@ -22,7 +22,7 @@ export class ResourceService extends BaseService {
     this.log(`Listing deployments for resource group '${this.resourceGroup}':`);
     const deployments = await this.resourceClient.deployments.listByResourceGroup(this.resourceGroup);
     return deployments.sort((a: DeploymentExtended, b: DeploymentExtended) => {
-      return (a.properties.timestamp > b.properties.timestamp) ? 1 : -1
+      return (a.properties.timestamp < b.properties.timestamp) ? 1 : -1
```
tbarlow12: 😨

src/services/resourceService.test.ts
```diff
@@ -102,10 +105,10 @@ describe("Resource Service", () => {
     const service = new ResourceService(sls, options);
     const deploymentString = await service.listDeployments();
     let expectedDeploymentString = "\n\nDeployments";
-    const originalTimestamp = +MockFactory.createTestTimestamp();
+    const originalTimestamp = +MockFactory.createTestTimestamp() + 4;
```
mydiemho: What's the `+` in front of MockFactory?

src/services/resourceService.test.ts
```diff
@@ -102,10 +105,10 @@ describe("Resource Service", () => {
...
     let i = 0
-    for (const dep of deployments) {
-      const timestamp = originalTimestamp + i
+    for (const dep of [...deployments].reverse()) {
```
mydiemho: how is this being reverse? what's the comparison operator?

src/services/resourceService.test.ts
tbarlow12: Parse as a number

src/services/resourceService.test.ts
tbarlow12: The original `deployments` object is in ascending order by timestamp. We want to make sure the deployments come back in descending order by timestamp.

The comparison operator isn't in the tests, it's in the actual code - `resourceService` line 25

src/services/resourceService.test.ts
mydiemho: magic :p

src/armTemplates/resources/apim.ts
```diff
@@ -61,7 +61,7 @@ export class ApimResource implements ArmResourceTemplateGenerator {
           "location": "[parameters('location')]",
           "sku": {
             "name": "[parameters('apimSkuName')]",
-            "capacity": "[parameters('apimCapacity')]"
+            "capacity": "[parameters('apimSkuCapacity')]"
```
wbreza: Nice catch!

src/services/resourceService.test.ts
wbreza: woah - really?? Is that a new typescript thing?

wbreza: Nit: Missing semicolon at end of line.

src/services/resourceService.test.ts
```diff
@@ -102,10 +105,10 @@ describe("Resource Service", () => {
...
+      const timestamp = originalTimestamp - i
```
wbreza: Nit: Missing semicolon at end of line.

PR #358, opened 2019-10-07
feat: Resource group tagging

tbarlow12:
## What did you implement:

Tagging for the Azure resource group as defined within the `provider`. Example

```yaml
...
provider:
    tags:
        TAG_1: tag1Value
        TAG_2: tag2Value
```

## How did you implement it:

- Check for existing resource group
- Merge tags from existing with tags from provider config
- Add tags to `createOrUpdate` call of resource group

## How can we verify it:

- Deploy a service with `tags` defined in provider config

OR

- Manually define tag for resource group in Azure portal
- Deploy service
- Tag will not be lost

## Todos:

       **Validate via `npm test`**
- [x] Write documentation

### DIFF
--- a/docs/DEPLOY.md
+++ b/docs/DEPLOY.md
@@ -14,6 +14,20 @@ By default, all functions are deploy to a stage (`dev`) and a region. The autoge
 - Create/Update app service plan, app service
 - Zipdeploy code
 
+### Tagging resource group
+
+To apply tags to your resource group, you can specify them in your provider config as such:
+
+```yaml
+...
+provider:
+  tags:
+    TAG_1: tagValue1
+    TAG_2: tagValue2
+```
+
+Existing tags with the same names will be updated. Existing tags (with different names) will **not** be removed.
+
 ### No resource group specified
 
 `sls deploy -s dev -r westus2`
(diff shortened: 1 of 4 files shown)


docs/DEPLOY.md
```diff
@@ -41,7 +55,7 @@ then user try to deploy
 `sls deploy -s prod -r westus`
 
 1. Do we use the environment variables / creds associated with `dev` or `prod`?
-1. Someone new to the codecase see both the `serverless.yaml` and the command in a CD script, how can they tell which resourceGroup goes where.
+2. Someone new to the codecase see both the `serverless.yaml` and the command in a CD script, how can they tell which resourceGroup goes where.
```
mydiemho: you should use follow markdown syntax and use `1.`.  It will auto-number on render

tbarlow12: Do you have some docs that specify that just `1.` is preferred? In the [markdown style guide on ordered lists](https://www.markdownguide.org/basic-syntax/#ordered-lists), it includes that as an _option_, but not the first one (another option being to just list any number at random, which will render to the same thing). Also, VSCode does an auto-complete to increment the number in markdown, which makes me think the explicit number is preferred (or at least commonly used). At the end of the day, they all render to the same thing, I was just curious if you had another source that stated that `1.` is preferred. Thanks for the feedback 😃

wbreza: You could use fibonacci sequence - it doesn't matter... will still come out as 1,2,3,4
1. a
3. b
5. c
8. d
13. e

PR #359, opened 2019-10-07
feat: Linux Runtime Support

tbarlow12:
## What did you implement:

- Generation of ARM template and code for deployment of linux Function apps.

## How did you implement it:

- Updated the arm templates to conditionally include values necessary for linux runtime function apps
- Force all linux function apps to run from external package, force all windows function apps to **not** run from external package

## How can we verify it:

Include following configuration in a function app deployment:

```yaml
provider:
    os: linux
```

## Todos:

       **Validate via `npm test`**
- [x] Write documentation

### DIFF
--- a/src/armTemplates/resources/appServicePlan.ts
+++ b/src/armTemplates/resources/appServicePlan.ts
@@ -1,9 +1,10 @@
 import { ArmResourceTemplate, ArmResourceTemplateGenerator, ArmParamType, ArmParameters, DefaultArmParams, ArmParameter } from "../../models/armTemplates";
-import { ResourceConfig, ServerlessAzureConfig } from "../../models/serverless";
+import { ResourceConfig, ServerlessAzureConfig, FunctionAppOS } from "../../models/serverless";
 import { AzureNamingService, AzureNamingServiceOptions } from "../../services/namingService";
 
 interface AppServicePlanParams extends DefaultArmParams {
   appServicePlanName: ArmParameter;
+  kind: ArmParameter;
   appServicePlanSkuName: ArmParameter;
   appServicePlanSkuTier: ArmParameter;
 }
@@ -23,6 +24,10 @@ export class AppServicePlanResource implements ArmResourceTemplateGenerator {
       appServicePlanName: {
         defaultValue: "",
         type: ArmParamType.String
+      },
+      kind: {
+        defaultValue: "",
+        type: ArmParamType.String,
       },
       location: {
         defaultValue: "",
@@ -49,6 +54,7 @@ export class AppServicePlanResource implements ArmResourceTemplateGenerator {
           "name": "[parameters('appServicePlanName')]",
           "type": "Microsoft.Web/serverfarms",
           "location": "[parameters('location')]",
+          "kind": "[parameters('kind')]",
           "properties": {
             "name": "[parameters('appServicePlanName')]",
             "workerSizeId": "3",
@@ -71,9 +77,14 @@ export class AppServicePlanResource implements ArmResourceTemplateGenerator {
       ...config.provider.storageAccount,
     };
 
+    const { os } = config.provider;
+
     const params: AppServicePlanParams = {
       appServicePlanName: {
         value: AppServicePlanResource.getResourceName(config),
+      },
+      kind: {
+        value: (os === FunctionAppOS.LINUX) ? "Linux" : undefined,
       },
       appServicePlanSkuName: {
         value: resourceConfig.sku.name,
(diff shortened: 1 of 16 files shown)


src/armTemplates/resources/appServicePlan.ts
```diff
@@ -24,6 +25,10 @@ export class AppServicePlanResource implements ArmResourceTemplateGenerator {
         defaultValue: "",
         type: ArmParamType.String
       },
+      kind: {
```
mydiemho: why the name `kind`?

src/armTemplates/resources/functionApp.ts
```diff
@@ -129,8 +147,8 @@ export class FunctionAppResource implements ArmResourceTemplateGenerator {
     const resourceConfig: FunctionAppConfig = {
       ...config.provider.functionApp,
     };
-    const { functionRuntime } = config.provider;
-
+    const { functionRuntime, os } = config.provider;
+    const linux = os === FunctionAppOS.LINUX;
```
mydiemho: can we add some parenthesis?

src/config.ts
```diff
@@ -40,7 +42,17 @@ export const configConstants = {
     stage: "dev",
     prefix: "sls",
     localPort: 7071,
+    os: FunctionAppOS.WINDOWS,
   },
+  dockerImages: {
+    node: {
+      "8": "DOCKER|microsoft/azure-functions-node8:2.0",
+      "10": "DOCKER|microsoft/azure-functions/node:2.0"
```
mydiemho: is this suppose to be `azure-functions-node`?

src/config.ts
```diff
@@ -40,7 +42,17 @@ export const configConstants = {
     stage: "dev",
     prefix: "sls",
     localPort: 7071,
+    os: FunctionAppOS.WINDOWS,
   },
+  dockerImages: {
```
mydiemho: why are these config under `dockerImages`?

src/services/configService.ts
```diff
@@ -178,28 +190,24 @@ export class ConfigService {
...
       appId: this.getOption(constants.variableKeys.appId)
         || process.env.AZURE_CLIENT_ID
-        || appId
+        || appId,
     }
     config.provider.resourceGroup = (
       this.getOption("resourceGroup", config.provider.resourceGroup)
     ) || AzureNamingService.getResourceName(options);
 
-    // Get property from `runFromBlobUrl` for backwards compatability
```
mydiemho: Is this no longer required for backwards compatibility?

src/config.ts
tbarlow12: No, that's the right name... For some reason they decided to change the naming convention that they used for the image when they moved to node 10... Had to look it up on docker hub because it's nowhere in the docs.

src/config.ts
tbarlow12: That's where we've been keeping any configuration-related constants. Linux function apps are built off of docker images that correspond to the configured runtime

src/services/configService.ts
tbarlow12: No, the tool is going to take the opinion that windows should run from package and Linux can only be external. It's not going to break anything, since Linux hasn't existed yet and running Windows from external package causes significant cold start and is not recommended anyway.

src/armTemplates/resources/functionApp.ts
tbarlow12: Parentheses would seem redundant to me, as there is only one operation happening here. Just capturing the boolean in a variable.

src/armTemplates/resources/appServicePlan.ts
tbarlow12: That's just the ARM specification required. See [the docs](https://docs.microsoft.com/en-us/azure/azure-functions/functions-infrastructure-as-code) for more details.

src/armTemplates/resources/functionApp.ts
mydiemho: I was actually confused about what this does until you explain it’s about capturing Boolean in a variable, so I don’t think it’s that redundant

src/services/armService.ts
```diff
@@ -231,7 +231,7 @@ export class ArmService extends BaseService {
 
       // This is a json path expression
       // Learn more @ https://goessner.net/articles/JsonPath/index.html#e2
-      const appSettingsPath = "$.resources[?(@.kind==\"functionapp\")].properties.siteConfig.appSettings";
+      const appSettingsPath = "$.resources[?(@.type==\"Microsoft.Web/sites\")].properties.siteConfig.appSettings";
```
mydiemho: Why is this changing?

src/armTemplates/resources/functionApp.ts
tbarlow12: That makes sense, I can be more expressive in the naming. I changed it to `isLinuxRuntime` to be more explicit.

I still don't think parentheses would be appropriate, since they're more for defining order of operations, and this is only one operation in the statement. Just like you wouldn't do `const a = (b + c);`

src/services/armService.ts
tbarlow12: This is a `jsonpath` query, to find the function app settings. The `kind` property was being used in the search, but it changed, so this search query is no longer valid. Now using the `type` property since that is still valid and unchanging.

PR #362, opened 2019-10-11
feat: Linux & Python Runtime Support

tbarlow12:
## What did you implement:

Runtime support for Linux and Python

## How did you implement it:

- Added `kind` to App Service plan and Function app ARM templates
- Added `reserved` and `linuxFxVersion` to Function App template
- Moved `spawn` to Utils class
- Added `CoreToolsService` to handle any CLI invocation of `func`
- Using `CoreToolsService` to package & publish Linux apps
- Only cleaning up built files when not publishing to Linux (see note below)
- Forces Python to run on Linux OS
- Use `/` instead of OS-specific path separator for `function.json` file (works for both windows and linux)
- Cleaning up generated python files in package service
- Fixed typo in name `uploadZippedArfifactToFunctionApp`

*Important*: Because core tools does not currently support publishing an existing package, any Linux applications will not clean built files. Also, because `func azure` (used in publishing) invokes the Azure CLI, I have not been able to get it to run properly as a child process. This will currently give the user the exact command to both publish their function app and then clean up the generated files.

## How can we verify it:

Specify the `linux` os in your `serverless.yml` like:

```yaml
provider:
    os: linux
```

and run `sls deploy`.

## Todos:

       **Validate via `npm test`**
- [x] Write documentation

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "typings": "./lib/index.d.ts",
   "author": "Azure Functions",
   "scripts": {
-    "start": "jest --watch",
+    "start": "npm run test -- --watch",
     "lint": "eslint src/**/*.ts",
     "lint:fix": "eslint src/**/*.ts --fix",
     "pretest": "npm run lint",
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "typings": "./lib/index.d.ts",
   "author": "Azure Functions",
   "scripts": {
-    "start": "npm run test -- --watch",
+    "start": "jest --watch",
     "lint": "eslint src/**/*.ts",
     "lint:fix": "eslint src/**/*.ts --fix",
     "pretest": "npm run lint",
--- a/src/armTemplates/resources/appServicePlan.ts
+++ b/src/armTemplates/resources/appServicePlan.ts
@@ -104,6 +104,11 @@ export class AppServicePlanResource implements ArmResourceTemplateGenerator {
       appServicePlanName: {
         value: AppServicePlanResource.getResourceName(config),
       },
+      /**
+       * `kind` only required for Linux Function Apps
+       * `undefined` values get removed from parameters by armService
+       * before deployment
+       */
       kind: {
         value: (os === FunctionAppOS.LINUX) ? "Linux" : undefined,
       },
--- a/src/armTemplates/resources/appServicePlan.ts
+++ b/src/armTemplates/resources/appServicePlan.ts
@@ -1,9 +1,10 @@
 import { ArmResourceTemplate, ArmResourceTemplateGenerator, ArmParamType, ArmParameters, DefaultArmParams, ArmParameter } from "../../models/armTemplates";
-import { ResourceConfig, ServerlessAzureConfig } from "../../models/serverless";
+import { ResourceConfig, ServerlessAzureConfig, FunctionAppOS } from "../../models/serverless";
 import { AzureNamingService, AzureNamingServiceOptions } from "../../services/namingService";
 
 interface AppServicePlanParams extends DefaultArmParams {
   appServicePlanName: ArmParameter;
+  kind: ArmParameter;
   appServicePlanSkuName: ArmParameter;
   appServicePlanSkuTier: ArmParameter;
   appServicePlanWorkerSizeId: ArmParameter;
@@ -27,6 +28,10 @@ export class AppServicePlanResource implements ArmResourceTemplateGenerator {
       appServicePlanName: {
         defaultValue: "",
         type: ArmParamType.String
+      },
+      kind: {
+        defaultValue: "",
+        type: ArmParamType.String,
       },
       location: {
         defaultValue: "",
@@ -69,6 +74,7 @@ export class AppServicePlanResource implements ArmResourceTemplateGenerator {
           "name": "[parameters('appServicePlanName')]",
           "type": "Microsoft.Web/serverfarms",
           "location": "[parameters('location')]",
+          "kind": "[parameters('kind')]",
           "properties": {
             "name": "[parameters('appServicePlanName')]",
             "workerSizeId": "[parameters('appServicePlanWorkerSizeId')]",
@@ -92,9 +98,14 @@ export class AppServicePlanResource implements ArmResourceTemplateGenerator {
       ...config.provider.appServicePlan,
     };
 
+    const { os } = config.provider;
+
     const params: AppServicePlanParams = {
       appServicePlanName: {
         value: AppServicePlanResource.getResourceName(config),
+      },
+      kind: {
+        value: (os === FunctionAppOS.LINUX) ? "Linux" : undefined,
       },
       appServicePlanSkuName: {
         value: resourceConfig.sku.name,
--- a/src/armTemplates/resources/functionApp.test.ts
+++ b/src/armTemplates/resources/functionApp.test.ts
@@ -1,27 +1,31 @@
 import { FunctionAppResource } from "./functionApp";
-import { ServerlessAzureConfig } from "../../models/serverless";
+import { ServerlessAzureConfig, FunctionAppOS, SupportedRuntimeLanguage } from "../../models/serverless";
 
 describe("Function App Resource", () => {
+
   const resourceGroupName = "myResourceGroup";
   const prefix = "prefix";
   const region = "westus";
   const stage = "prod";
 
-  it("generates the correct resource name", () => {
-    const config: ServerlessAzureConfig = {
-      provider: {
-        name: "azure",
-        prefix,
-        region,
-        stage,
-        resourceGroup: resourceGroupName,
-        runtime: "nodejs10.x"
-      },
-      service: ""
-    } as any;
+  const defaultConfig: ServerlessAzureConfig = {
+    provider: {
+      name: "azure",
+      prefix,
+      region,
+      stage,
+      resourceGroup: resourceGroupName,
+      functionRuntime: {
+        language: SupportedRuntimeLanguage.NODE,
+        version: "10.6.0"
+      }
+    },
+    service: ""
+  } as any;
 
-    expect(FunctionAppResource.getResourceName(config)).toEqual(
-      `${config.provider.prefix}-wus-${config.provider.stage}`
+  it("generates the correct resource name", () => {
+    expect(FunctionAppResource.getResourceName(defaultConfig)).toEqual(
+      `${defaultConfig.provider.prefix}-wus-${defaultConfig.provider.stage}`
     );
   });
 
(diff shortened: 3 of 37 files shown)


package.json
```diff
@@ -7,7 +7,7 @@
   "typings": "./lib/index.d.ts",
   "author": "Azure Functions",
   "scripts": {
-    "start": "npm run test -- --watch",
+    "start": "jest --watch",
```
wbreza: why the change to trigger jest directly vs. the NPM command.  It doesn't seem like a big change but now `pretest` etc are also not run which did some linting, etc.

src/armTemplates/resources/functionApp.test.ts
```diff
@@ -1,51 +1,126 @@
...
 
+  const defaultConfig: ServerlessAzureConfig = {
+    provider: {
+      name: "azure",
+      prefix,
+      region,
+      stage,
+      resourceGroup: resourceGroupName,
+      functionRuntime: {
+        language: SupportedRuntimeLanguage.NODE,
```
wbreza: Does this interfere with setting the os / runtime directly on the provider?  What are we recommending to users?  Why do we have 2 ways?

src/armTemplates/resources/functionApp.test.ts
```diff
@@ -1,51 +1,126 @@
...
+    });
+  
+    it("gets correct linux parameters - node 8", () => {
+      const config: ServerlessAzureConfig = {
+        ...defaultConfig,
+        provider: {
+          ...defaultConfig.provider,
+          os: FunctionAppOS.LINUX,
+          functionRuntime: {
+            language: SupportedRuntimeLanguage.NODE,
```
wbreza: Same as above - how do I know what param to set in the yaml?

src/armTemplates/resources/functionApp.ts
```diff
@@ -39,6 +43,18 @@ export class FunctionAppResource implements ArmResourceTemplateGenerator {
         defaultValue: "",
         type: ArmParamType.String
       },
+      functionAppKind: {
+        defaultValue: "functionapp",
+        type: ArmParamType.String,
+      },
+      functionAppReserved: {
+        defaultValue: false,
```
wbreza: We should add some inline comments in the ARM template resources that describe the purpose of each of these properties.

src/config.ts
```diff
@@ -40,7 +46,28 @@ export const configConstants = {
...
+  dockerImages: {
+    node: {
+      "8": "DOCKER|microsoft/azure-functions-node8:2.0",
+      "10": "DOCKER|microsoft/azure-functions/node:2.0"
+    },
+    python: {
+      "3.6": "DOCKER|microsoft/azure-functions/python:2.0"
+    }
   },
+  defaultFuncIgnore: [
```
wbreza: Does this play nice with Node apps on linux?  If I don't use webpack I would need to ensure my `package.json` and `node_modules` are included in my deployment.

src/plugins/azureBasePlugin.ts
```diff
@@ -10,6 +10,10 @@ export abstract class AzureBasePlugin<TOptions=Serverless.Options> {
   protected config: ServerlessAzureConfig;
   protected commands: ServerlessCommandMap;
   protected processedCommands: ServerlessCliCommand[];
+  /**
+   * Specifies if targeted runtime is linux
+   */
+  protected isLinuxTarget: boolean;
```
wbreza: Doesn't feel like this property should exist on the base plugin since its a concern of deployment

src/plugins/login/utils/simpleFileTokenCache.test.ts
```diff
@@ -1,9 +1,12 @@
-import fs from "fs";
 import path from "path";
 import os from "os";
 import mockFs from "mock-fs";
 import { MockFactory } from "../../../test/mockFactory";
 import { SimpleFileTokenCache } from "./simpleFileTokenCache";
+let fs;
+jest.isolateModules(() => {
```
wbreza: What does the `isolateModules()` do?

wbreza: Is it needed since you figured out the mock/restore stuff?

src/services/baseService.test.ts
```diff
@@ -1,10 +1,14 @@
-import fs from "fs";
 import mockFs from "mock-fs";
 import Serverless from "serverless";
 import { ServerlessAzureOptions } from "../models/serverless";
 import { MockFactory } from "../test/mockFactory";
 import { BaseService } from "./baseService";
 
+let fs;
```
wbreza: Confirm if this complexity is needed.

src/services/baseService.ts
```diff
@@ -38,6 +41,9 @@ export abstract class BaseService {
     this.resourceGroup = this.configService.getResourceGroupName();
     this.deploymentName = this.configService.getDeploymentName();
     this.artifactName = this.configService.getArtifactName(this.deploymentName);
+    this.runtime = this.configService.getRuntime();
+    this.pythonTarget = this.runtime.language === SupportedRuntimeLanguage.PYTHON;
```
wbreza: Similar to above - doesn't feel like python/linux related variables should exist on the base service.  Also, since they are boolean variables i would rename to something like `isPythonTarget` && `isLinuxTarget`

src/services/functionAppService.ts
```diff
@@ -157,51 +158,19 @@ export class FunctionAppService extends BaseService {
...
-        functionApp,
-        configConstants.runFromPackageSetting,
-        sasUrl
-      );
-
-      await this.syncTriggers(functionApp, response.properties);
+    // Uploaded to blob storage as artifact for future reference
+    await this.uploadZippedArtifactToBlobStorage(functionZipFile);
+
+    if (this.linuxTarget) {
```
wbreza: This is better - but I think this block should be encapsulated in a separate service so that the `functionAppService` doesn't need to care about the concerns of the OS.

src/services/functionAppService.ts
```diff
@@ -223,7 +192,7 @@ export class FunctionAppService extends BaseService {
     return await this.get();
   }
 
-  public async uploadZippedArfifactToFunctionApp(functionApp: Site, functionZipFile: string) {
```
wbreza: Arfifact 🤣

src/services/packageService.test.ts
```diff
@@ -241,7 +247,97 @@ describe("Package Service", () => {
...
     copyFileSpy.mockRestore();
   });
+
+  describe("Python packages", () => {
+    let mySpawn;
+    let packageService: PackageService;
+    let functions: string[];
+
+    beforeEach(() => {
+
```
wbreza: nit: empty lines

src/services/packageService.ts
```diff
@@ -69,30 +88,80 @@ export class PackageService extends BaseService {
...
   public createBinding(functionName: string, functionMetadata: FunctionMetadata) {
-    const functionJSON = functionMetadata.params.functionsJson;
-    functionJSON.entryPoint = functionMetadata.entryPoint;
-    functionJSON.scriptFile = functionMetadata.handlerPath;
+    const functionJSON = this.getFunctionJson(functionName, functionMetadata);
+    const functionDirPath = this.makeFunctionDir(functionName);
+
+    fs.writeFileSync(path.join(functionDirPath, "function.json"), this.stringify(functionJSON));
+
+    if (this.runtime.language === SupportedRuntimeLanguage.PYTHON) {
```
wbreza: What is happening here?  Writing an empty `__init__.py` file?  How does this work?  Would be good to have some comments here with intention of this change.

package.json
tbarlow12: Was going to ask your thoughts on this. I thought it was kind of nice to have a command to just trigger tests without having to wait for linting to finish. It was taking kind of a while for me, but I'm open to leaving this as is

src/armTemplates/resources/functionApp.test.ts
tbarlow12: This is an object that I'm inferring from the `runtime` in the provider, and then setting in `functionRuntime`. I didn't want to change `runtime` directly, but I can if that would be better. I just didn't want to parse and instantiate this object every time

src/armTemplates/resources/functionApp.test.ts
tbarlow12: User would only set `runtime`

src/config.ts
tbarlow12: I'll update this.

src/plugins/azureBasePlugin.ts
tbarlow12: I think this will go away when I address your other feedback on reducing some of the complexity

src/plugins/login/utils/simpleFileTokenCache.test.ts
tbarlow12: Nope. Good catch. I'll remove

src/services/packageService.ts
tbarlow12: I'll see if it works without it, but I've always had to create `__init__.py` in order for the directory to be considered a package. [Helpful article for more info](https://medium.com/@ramrajchandradevan/python-init-py-modular-imports-81b746e58aae#targetText=The%20__init__.py,on%20the%20module%20search%20path.)

src/armTemplates/resources/functionApp.test.ts
tbarlow12: Will add as private configuration within config service

src/armTemplates/resources/functionApp.test.ts
tbarlow12: Same as above

src/plugins/azureBasePlugin.ts
tbarlow12: Add to config service

src/services/functionAppService.ts
tbarlow12: Will create a `functionAppPublishService`

src/armTemplates/resources/functionApp.test.ts
tbarlow12: I just realized that I can't do this ^. It needs to be inside the `ServerlessAzureConfig` because that's the only data that is passed to the ARM template resources like `functionApp`. Will add a comment on the property indicating that it's only for internal use

src/services/packageService.ts
tbarlow12: Removed this

src/armTemplates/resources/appServicePlan.ts
```diff
@@ -71,10 +77,15 @@ export class AppServicePlanResource implements ArmResourceTemplateGenerator {
...
     };
 
+    const { os } = config.provider;
+
     const params: AppServicePlanParams = {
       appServicePlanName: {
         value: AppServicePlanResource.getResourceName(config),
       },
+      kind: {
+        value: (os === FunctionAppOS.LINUX) ? "Linux" : undefined,
```
pjlittle: interesting - so undefined implies Windows?

tbarlow12: The armService removes parameters with undefined parameters on deployment. So by keeping it undefined, it just won't be added

src/shared/utils.ts
```diff
@@ -188,4 +200,54 @@ export class Utils {
...
+    }
+
+    const env = {
+      // Inherit environment from current process, most importantly, the PATH
+      ...process.env,
+      // Environment variables from serverless config are king
+      ...serverless.service.provider["environment"],
+    }
+    if (!options.silent) {
+      serverless.cli.log(`Spawning process '${command} ${commandArgs.join(" ")}'`);
```
mydiemho: nit: we should refactor the log method to `log`, `info`, etc. and turn each on/off accordingly with the `silent` flag

src/armTemplates/resources/functionApp.ts
```diff
@@ -1,14 +1,52 @@
...
+import { FunctionAppConfig, ServerlessAzureConfig, SupportedRuntimeLanguage, FunctionAppOS } from "../../models/serverless";
 import { AzureNamingService, AzureNamingServiceOptions } from "../../services/namingService";
+import configConstants from "../../config";
 
 interface FunctionAppParams extends DefaultArmParams {
+  /**
+   * Name of function app
+   */
   functionAppName: ArmParameter;
+  /**
```
mydiemho: nit: new line before comments for readability, especially for text to speech reader

src/armTemplates/resources/functionApp.ts
```diff
@@ -1,14 +1,52 @@
...
+   * Node version of function app
+   */
   functionAppNodeVersion: ArmParameter;
+  /**
```
mydiemho: same for all the interface elements below

src/shared/utils.ts
tbarlow12: Yep, that's already in the [backlog](https://github.com/serverless/serverless-azure-functions/issues/374)

tbarlow12: Although this `log` method is part of the serverless core, not something we can refactor.

src/armTemplates/resources/functionApp.ts
tbarlow12: Hmm, not sure. We haven't done that before, and we don't have new lines between the members of other interfaces. I don't know if a jsdoc comment should change that

Vbubblery:
Hello,

May I ask when could you merge this PR?

src/config.ts
```diff
@@ -40,8 +46,26 @@ export const configConstants = {
...
+  },
+  dockerImages: {
+    node: {
+      "8": "DOCKER|microsoft/azure-functions-node8:2.0",
+      "10": "DOCKER|microsoft/azure-functions/node:2.0"
+    },
+    python: {
+      "3.6": "DOCKER|microsoft/azure-functions/python:2.0"
+    }
   },
```
Hazhzeng: In Linux, currently we accepts `PYTHON|3.6`, `PYTHON|3.7`, `PYTHON|3.8`, `NODE|8`, `NODE|10` in the **LinuxFxVersion**, and we will convert these settings to the proper images for you.

Placing a list of MCR docker images have been deprecated in LinuxFxVersion setting.

tbarlow12:
Closing in favor of coming PR

PR #363, opened 2019-10-15
fix: Publish Service to handle Function App publishing

tbarlow12:
This is a PR addressing @wbreza's feedback on [previous PR](https://github.com/serverless/serverless-azure-functions/pull/362). I figured the changes were big enough to separate the 2 and make it easier to see.

### DIFF
--- a/src/plugins/package/azurePackagePlugin.ts
+++ b/src/plugins/package/azurePackagePlugin.ts
@@ -1,6 +1,6 @@
 
 import Serverless from "serverless";
-import { ServerlessCliCommand } from "../../models/serverless";
+import { ServerlessCliCommand, FunctionAppOS } from "../../models/serverless";
 import AzureProvider from "../../provider/azureProvider";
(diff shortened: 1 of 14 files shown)


src/plugins/package/azurePackagePlugin.ts
```diff
@@ -1,13 +1,14 @@
...
-import { ServerlessCliCommand } from "../../models/serverless";
+import { ServerlessCliCommand, FunctionAppOS } from "../../models/serverless";
 import AzureProvider from "../../provider/azureProvider";
 import { PackageService } from "../../services/packageService";
 import { AzureBasePlugin } from "../azureBasePlugin";
 
 export class AzurePackagePlugin extends AzureBasePlugin {
   private bindingsCreated: boolean = false;
   public provider: AzureProvider;
+  private isLinuxTarget: boolean;
```
tbarlow12: This is used in multiple hooks. Removed from base plugin, but I still think it makes sense to have here.

PR #383, opened 2019-10-17
feat: Prompt user for confirmation to remove resource group

### LINKED ISSUE #380: Remove command should prompt for confirmation
# This is a Feature Proposal

## Description

- The `remove` command deletes the entire resource group, which is non-recoverable
- The plugin should at least prompt for confirmation upon running that command (maybe typing the resource group name, just like the Azure portal?)

Similar or dependent issues:

- #356

tbarlow12:
## What did you implement:

Requires user to enter resource group name as confirmation for group removal

Closes #380 

## How did you implement it:

- Added utility function to wait for user input
- Check user input against configured resource group name
- Delete if results match

## How can we verify it:

- Try to remove a non-deployed project
- Run `sls remove` and then just type `Enter` (empty string shouldn't match), removal won't happen
- Run `sls remove` and then type exact resource group name and removal will happen

## Todos:

       **Validate via `npm test`**
- [x] Write documentation

### DIFF
--- a/src/plugins/remove/azureRemovePlugin.ts
+++ b/src/plugins/remove/azureRemovePlugin.ts
@@ -48,7 +48,8 @@ export class AzureRemovePlugin extends AzureBasePlugin {
       this.log(`Resource group "${rgName}" does not exist in your Azure subscription`)
       return;
     }
-    this.log(`This command will delete your entire resource group (${resourceService.getResourceGroupName()}). ` +
+    this.log(`This command will delete your ENTIRE resource group (${resourceService.getResourceGroupName()}). ` +
+      "and ALL the Azure resources that it contains " +
       "Are you sure you want to proceed? If so, enter the full name of the resource group :");
     const input = await Utils.waitForUserInput();
     if (input === resourceService.getResourceGroupName()) {
--- a/src/plugins/remove/azureRemovePlugin.ts
+++ b/src/plugins/remove/azureRemovePlugin.ts
@@ -1,6 +1,7 @@
 import Serverless from "serverless";
 import { ResourceService } from "../../services/resourceService";
 import { AzureBasePlugin } from "../azureBasePlugin";
+import { Utils } from "../../shared/utils";
 
 export class AzureRemovePlugin extends AzureBasePlugin {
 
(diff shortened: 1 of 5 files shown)


src/plugins/remove/azureRemovePlugin.ts
```diff
@@ -40,10 +41,25 @@ export class AzureRemovePlugin extends AzureBasePlugin {
...
+    if (!rg) {
+      this.log(`Resource group "${rgName}" does not exist in your Azure subscription`)
+      return;
+    }
+    this.log(`This command will delete your entire resource group (${resourceService.getResourceGroupName()}). ` +
+      "Are you sure you want to proceed? If so, enter the full name of the resource group :");
+    const input = await Utils.waitForUserInput();
+    if (input === resourceService.getResourceGroupName()) {
+      this.log("Deleting resource group");
+      await resourceService.deleteDeployment();
```
wbreza: What does `deleteDeployment()` do?  Do we even need it if we are removing the whole resource group?

tbarlow12: I was actually wondering the same thing... I don't remember implementing this (although it's possible that I still did), but I didn't want to change the functionality in this PR without finding out why it was there

src/plugins/remove/azureRemovePlugin.ts
```diff
@@ -40,10 +41,25 @@ export class AzureRemovePlugin extends AzureBasePlugin {
...
-
-    this.log("Service successfully removed");
+    const resourceService = new ResourceService(this.serverless, this.options);
+    const rg = await resourceService.getResourceGroup();
+    const rgName = resourceService.getResourceGroupName();
+    if (!rg) {
+      this.log(`Resource group "${rgName}" does not exist in your Azure subscription`)
+      return;
+    }
+    this.log(`This command will delete your entire resource group (${resourceService.getResourceGroupName()}). ` +
```
pjlittle: nit: can we update the message to indicate something like: "This command will delete your ENTIRE resource group <rg name> and ALL the Azure resources that it contains."

Just to explicitly call out that all resources will get nuked, not just things created by the sls framework.

pjlittle:
Minor comment, otherwise looks great. Thx for adding this!

PR #399, opened 2019-12-26
Fix: Use correct displayName for cosmosDBTrigger

### LINKED ISSUE #389: CosmosDB bindings is broken in 1.0.2-13
# This is a Bug Report

## Description

- What went wrong?
Got error when using `cosmosDB` binding 
```
  Error --------------------------------------------------
 
  Error: Binding not supported
```

- What did you expect should have happened?
- What was the config you used?
- What stacktrace or error message from your provider did you see?

Similar or dependent issues:

## Additional Data

- **_Serverless Framework Version you're using_**: 1.55.1
- **_Serverless CLI Version you're using_**: 1.4.0
- **_Serverless Azure Plugin Version you're using_**: 1.0.2-13
- **_Operating System_**: macOS
- **_Stack Trace_**:
- **_Provider Error messages_**:

ofekray:
## What did you implement:
Used correct displayName for cosmosDBTrigger to fix getting "Binding not supported" error.

Closes #389

## How did you implement it:

Changed the displayName from $cosmosDBIn_displayName to $cosmosDBTriggerIn_displayName

## How can we verify it:

Try to deploy the following function before & after the change:
```
  cosmosChangeFeed:
    handler: cosmosChangeFeed.handler
    events:
      - cosmosDB:
        x-azure-settings:
          direction: in
          name: documents
          databaseName: databaseName
          collectionName: containerName
          leaseCollectionName: leases
          connectionStringSetting: COSMOS_DB_CONNECTION
          createLeaseCollectionIfNotExists: true
```

### DIFF
--- a/src/shared/bindings.json
+++ b/src/shared/bindings.json
@@ -1305,7 +1305,7 @@
     },
     {
       "type": "cosmosDBTrigger",
-      "displayName": "$cosmosDBIn_displayName",
+      "displayName": "$cosmosDBTriggerIn_displayName",
       "direction": "trigger",
       "enabledInTryMode": false,
       "documentation": "$content=Documentation\\cosmosDBTrigger.md",


ofekray:
@tbarlow12, @wbreza.
This is a pretty small fix.
Would appreciate if you can take a look.

tbarlow12:
Going to close and re-open to see if the pipeline is working

tbarlow12:
@mydiemho It looks like the pipelines aren't running for external contributors? Could you take a look when you get some time?

tbarlow12:
@ofekray Sorry for the radio silence. We're investigating why the pipelines aren't running for external contributors. Also, this change will require some additional fixes because the plugin first tries to get the `bindings.json` file from the official azure functions repo, where we got this original file. We'll probably just need to update that so it either pins to the version we have or check with that team to see why their file is wrong.

mydiemho:
@tbarlow12  @ofekray the pipeline wasn't block for external contributor just access to a token.  It is now unblock but we just have to fix a different issues on our end

PR #401, opened 2019-12-30
feat: Logging Service to allow configurable verbosity

### LINKED ISSUE #374: Ability to set logging verbosity
# This is a Feature Proposal

## Description

- Currently, any and all logs are displayed every time, no matter what.
- This is a proposal to allow the user to define logging level of the plugin (even if it's just `--verbose` or not)
- Would need to implement a `LoggingService` that would be used both in the `BaseService` and the `BasePlugin` `log` functions.
- Would need to look through all existing log statements and decide what level of verbosity they should fall into
- Could add additional log statements for the most verbose setting

tbarlow12:
## What did you implement:

A `LoggingService` class that allows for configurable verbosity of logging

- To line up with the SLS core, the flag `--verbose` or `-v` (shortcut) with no value sets to most verbose (debug)
- Users can also set a specific verbosity by providing a value after the flag. For example `-v error`, `-v debug`

Closes #374 

## How did you implement it:

`LoggingService` is used by both the `BaseService` and the `BasePlugin`, which make all current logging calls. Added calls for `error`, `warn`, `info` and `debug` to both base service and base plugin. Did not update to use any of the new calls, everything is currently logged as `info`. Which means that if you were to set verbosity to either `warn` or `error` in the CLI, there would be no log statements shown

## How can we verify it:

Add `-v error` to any CLI run. There will be no log statements printed. Then try with `-v info` or `-v debug`, and there will be log statements printed. The codebase will need to be updated with the appropriate level of log for each current statement where necessary, but that is outside the scope of this PR.

## Todos:

       **Validate via `npm test`**
- [x] Write documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,6 +216,15 @@ The getting started walkthrough illustrates the interactive login experience, wh
 - **[Visit our sample repos](docs/examples/samples.md) for full projects with different use cases**
 - [Configuring API Management](docs/examples/apim.md) that sits in front of function app
 
+### Logging Verbosity
+
+You can set the logging verbosity with the `--verbose` flag (`-v` for short). If the `--verbose` flag is set with no value, logging will be as verbose as possible (debug mode). You can also provide a value with the flag to set the verbosity to a specific level:
+
+- `-v error` - Only error messages printed
+- `-v warn` - Only error and warning messages printed
+- `-v info` - Only error, warning and info messages printed
+- `-v debug` - All messages printed
+
 ### Contributing
 
 Please create issues in this repo for any problems or questions you find. Before sending a PR for any major changes, please create an issue to discuss.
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -216,6 +216,15 @@ The getting started walkthrough illustrates the interactive login experience, wh
 - **[Visit our sample repos](docs/examples/samples.md) for full projects with different use cases**
 - [Configuring API Management](docs/examples/apim.md) that sits in front of function app
 
+### Logging Verbosity
+
+You can set the logging verbosity with the `--verbose` flag (`-v` for short). If the `--verbose` flag is set with no value, logging will be as verbose as possible (debug mode). You can also provide a value with the flag to set the verbosity to a specific level:
```
PIC123: Is the default mode debug here? It seems that in the logging service, the default is 'info'

tbarlow12: Default is `info`. `debug` is set if the `--verbose` flag is passed with no argument, or if there is an argument, it sets it to that value

Issue #12, opened 2017-02-20
Discuss: need to use x-azure-settings?


christopheranderson:
One thing that has occurred to me as I use the tooling is that the x-azure-settings stuff and the way we do bindings is a bit odd. I looked at other providers and they don't bother with separating out their special settings - could we provide a more optimal experience by just skipping the "x-azure" stuff?

lostintangent:
+1! I would love to see the prefix removed as well.

tbarlow12:
Closing in favor of duplicate #371

PR #404, opened 2019-12-31
feat: Eliminate need for 'x-azure-settings' config

### LINKED ISSUE #371: Remove `x-azure-settings` from function configuration
# This is a Feature Proposal

## Description

- Currently, the configuration for a function includes `x-azure-settings` as a property, and any function binding metadata exists as a child to that property.
- This is a proposition to eliminate `x-azure-settings` from the function configuration and instead make its sub-settings children of the trigger-type name. For example, the configuration

```yml
functions:
  hello:
    handler: hello/hello.main
    events:
      - http: true
        x-azure-settings:
          methods:
            - GET
          authLevel: anonymous
```
would become:
```yml
functions:
  hello:
    handler: hello/hello.main
    events:
      - http:
          methods:
             - GET
          authLevel: anonymous
```
and the configuration:
```yml
storageBlob:
    handler: src/handlers/storageBlob
    events:
      - blob:
        x-azure-settings:
          name: blob
          path: blob-sample/{blobName}
          connection: AzureWebJobsStorage
```
would become:
```yml
storageBlob:
    handler: src/handlers/storageBlob
    events:
      - blob:
          name: blob
          path: blob-sample/{blobName}
          connection: AzureWebJobsStorage
```

This also would remove the `http: true` for HTTP triggers, which is inconsistent with the rest of the triggers. Its existence would imply `true`

Similar or dependent issues:

- #12 
- #37 
- #38

tbarlow12:
## What did you implement:

Users can now place the configuration in the event object rather than inside the nested `x-azure-settings` configuration. **Users can still use `x-azure-settings` if they want. This is NOT a breaking change**

Closes #371 and #12 

## How did you implement it:

- Made `x-azure-settings` to be optional member of interface
- To maintain backwards compatibility, the binding utilities will first check to see if `x-azure-functions` is defined on the event object. If it is not, the event itself is used.
- The Func plugin and service that generate the .yml file were also updated to reflect the change
- Changes to the MockFactory, tests, and documentation were made to reflect the change

## How can we verify it:

- Take any existing project, move the settings contained within `x-azure-settings` up one level and run `sls offline`. The bindings are generated correctly

## Todos:

       **Validate via `npm test`**
- [x] Write documentation

### DIFF
--- a/src/models/serverless.ts
+++ b/src/models/serverless.ts
@@ -156,7 +156,13 @@ export interface ServerlessAzureFunctionConfig {
 
 export interface ServerlessAzureFunctionBindingConfig {
   http?: boolean;
-  "x-azure-settings": ServerlessExtraAzureSettingsConfig;
+  direction?: string;
+  route?: string;
+  name?: string;
+  authLevel?: string;
+  methods?: string[];
+  /** Maintained for backwards compatibility */
+  "x-azure-settings"?: ServerlessExtraAzureSettingsConfig;
 }
 
 export interface ServerlessExtraAzureSettingsConfig {
--- a/src/shared/utils.ts
+++ b/src/shared/utils.ts
@@ -156,14 +156,14 @@ export class Utils {
 
   public static getIncomingBindingConfig(functionConfig: ServerlessAzureFunctionConfig) {
     return functionConfig.events.find((event) => {
-      const settings = event["x-azure-settings"]
+      const settings = Utils.get(event, constants.xAzureSettings, event);
       return settings && (!settings.direction || settings.direction === "in");
     });
   }
 
-  public static getOutgoingBinding(functionConfig: ServerlessAzureFunctionConfig) {
+  public static getOutgoingBindingConfig(functionConfig: ServerlessAzureFunctionConfig) {
     return functionConfig.events.find((event) => {
-      const settings = event["x-azure-settings"]
+      const settings = Utils.get(event, constants.xAzureSettings, event);
       return settings && settings.direction === "out";
     });
   }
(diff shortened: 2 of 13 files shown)


src/models/serverless.ts
```diff
@@ -167,7 +167,13 @@ export interface ServerlessAzureFunctionConfig {
...
 export interface ServerlessAzureFunctionBindingConfig {
   http?: boolean;
-  "x-azure-settings": ServerlessExtraAzureSettingsConfig;
+  direction?: string;
+  route?: string;
+  name?: string;
+  authLevel?: string;
+  methods?: string[];
+  /** Maintained for backwards compatibility */
+  "x-azure-settings"?: ServerlessExtraAzureSettingsConfig;
```
wbreza: Great to see that it is backwards compatible.

src/shared/utils.ts
```diff
@@ -143,14 +143,14 @@ export class Utils {
...
+      const settings = Utils.get(event, constants.xAzureSettings, event);
       return settings && (!settings.direction || settings.direction === "in");
     });
   }
 
-  public static getOutgoingBinding(functionConfig: ServerlessAzureFunctionConfig) {
+  public static getOutgoingBindingConfig(functionConfig: ServerlessAzureFunctionConfig) {
     return functionConfig.events.find((event) => {
-      const settings = event["x-azure-settings"]
+      const settings = Utils.get(event, constants.xAzureSettings, event);
```
wbreza: Consider reusing this constant in other places where the raw text is still duplicated.

src/test/mockFactory.ts
```diff
@@ -334,15 +334,15 @@ export class MockFactory {
     };
   }
 
-  public static createTestFunctionMetadata(name: string): ServerlessAzureFunctionConfig {
+  public static createTestFunctionMetadata(name: string, xAzureSettings: boolean = true): ServerlessAzureFunctionConfig {
```
wbreza: Seems a little messy to add a new param to all these mock methods special for the xAzureSettings case.  Maybe consider creating new mock method for testing?

PR #434, opened 2020-04-17
feat: Use cross-spawn instead of adding .cmd

tbarlow12:
## What did you implement:

Instead of manually checking for the operating system, this uses the well known package `cross-spawn` (142M downloads per month) to spawn the process across operating systems.

## How did you implement it:

Replaced the call to `child_process.spawn` with `cross-spawn.spawn`. Also added `jest-os-detection` that will only test the spawn calls the way that they should be expected on that native os rather than defining a property on the `process` object

## How can we verify it:

Run `sls offline`. It will spawn on any operating system. Run `npm test` on mac, linux or windows. You'll see that 8 tests are skipped on any OS (each OS has 4 tests)

## Todos:

       **Validate via `npm test`**
- [x] Write documentation

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -23,5 +23,6 @@ module.exports = {
     "json",
     "jsx",
     "node"
-  ]
+  ],
+  "setupFilesAfterEnv": ["jest-os-detection"]
 };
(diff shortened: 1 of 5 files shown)


pjlittle:
Looks good - much better way of handling OS matrix. 👍

PR #435, opened 2020-04-17
fix: Remove -v shortcut for --verbose, conflicts with --version

tbarlow12:
## What did you implement:

Currently, if you were to run `sls --version` or `sls -v`, the plugin causes it to fail because the `LoggingService` creates a shortcut for `--verbose`, which is `-v`. This conflicts with `--version`. This PR removes that shortcut and updates the documentation.

## How did you implement it:

Only look for the full `verbose` flag within the options, set log level to `info` if undefined.

## How can we verify it:

Run `sls --version`. It shouldn't fail.

## Todos:

       **Validate via `npm test`**
- [x] Write documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,12 +228,12 @@ The getting started walkthrough illustrates the interactive login experience, wh
 
 ### Logging Verbosity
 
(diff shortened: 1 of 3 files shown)


pjlittle:
LGTM - thanks for fixing, we definitely shouldn't be stomping/failing on this.

PR #457, opened 2020-05-18
ci: .NET Core Integration tests & Workflow fixes

tbarlow12:
## What did you implement:

Adds .NET core integration tests that will be triggered 2x daily as well as any pushes to branch with pattern `*/trigger-int-tests`

## How did you implement it:

- Added templates for .NET Core versions 3.1 & 2.2 for Linux and Windows
- Updated the workflow to create jobs for all multi-os runtimes, keeping Python in its own job

## How can we verify it:

Check out and push to a branch with pattern `*/trigger-int-tests` and go to `Actions` page on repo

## Todos:

       **Validate via `npm test`**
- [x] Write documentation


tbarlow12:
Just a heads up to reviewers: The only reason integration tests show in this PR is because the workflow is set to trigger on branches that follow the pattern `*/trigger-int-tests`. This was used to test the workflow, and will be used in the future to manually trigger the workflow if needed. They will not be part of PRs going forward

Issue #444, opened 2020-04-22
Cannot set property 'name' of undefined Python timer


jeicob22:
# This is a Bug Report

## Description

- What went wrong?
Error when running  "sls offline" and "sls deploy" for deploying a Python 3.7 Function with a "timer" event.

- What did you expect should have happened?
To Successfully deploy the timer/cron

- What was the config you used?
```yaml
service: megatest-num2

provider:
  name: azure
  location: West Europe
  region: westeurope
  runtime: "python3.7"
  os: "linux"
  resourceGroup: my-mega-test-rg
  storageAccount:
    httpsTrafficOnly: true

plugins: # look for additional plugins in the community plugins repo: https://github.com/serverless/plugins
  - serverless-azure-functions

# you can add packaging information here
package:
  include:
    - src/shared/**
  exclude:
    - local.settings.json
    - .vscode/**

functions:
  SyncDirUsers:
    handler: "src/handlers/SyncDirUsers.main"
    events:
      - timer: true
        x-azure-settings:
          name: myCron
          schedule: '* */10 * * * *'
```
src/handlers/SyncDirUsers.main

```python
import azure.functions as func
import logging

def main(mytimer: func.TimerRequest) -> None:
    logging.info('Python received a cron trigger.')
```

- What stacktrace or error message from your provider did you see?
```bash
Serverless: Load command interactiveCli
Serverless: Load command config
Serverless: Load command config:credentials
Serverless: Load command config:tabcompletion
Serverless: Load command config:tabcompletion:install
Serverless: Load command config:tabcompletion:uninstall
Serverless: Load command create
Serverless: Load command install
Serverless: Load command package
Serverless: Load command deploy
Serverless: Load command deploy:function
Serverless: Load command deploy:list
Serverless: Load command deploy:list:functions
Serverless: Load command invoke
Serverless: Load command invoke:local
Serverless: Load command info
Serverless: Load command logs
Serverless: Load command metrics
Serverless: Load command print
Serverless: Load command remove
Serverless: Load command rollback
Serverless: Load command rollback:function
Serverless: Load command slstats
Serverless: Load command plugin
Serverless: Load command plugin
Serverless: Load command plugin:install
Serverless: Load command plugin
Serverless: Load command plugin:uninstall
Serverless: Load command plugin
Serverless: Load command plugin:list
Serverless: Load command plugin
Serverless: Load command plugin:search
Serverless: Load command config
Serverless: Load command config:credentials
[...]

PR #480, opened 2020-06-29
fix: Check for valid index before setting name property

mikedingjan:
This commit fixes a bug when using the timer trigger where it tries to set the name of the binding. The name of the binding can't be set since there is no index found (-1)

Fixes: #444

### DIFF
--- a/src/services/packageService.test.ts
+++ b/src/services/packageService.test.ts
@@ -5,6 +5,7 @@ import Serverless from "serverless";
 import { FunctionMetadata } from "../shared/utils";
 import { MockFactory } from "../test/mockFactory";
 import { PackageService } from "./packageService";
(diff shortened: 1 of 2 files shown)


tbarlow12:
@mikedingjan thanks for the fix! Any way we can get a unit test or two to validate the change?

mikedebock:
@tbarlow12 Thank you for your feedback. I'm also waiting for this fix, because of that I added the unittest :)

tbarlow12:
@mikedingjan and @mikedebock thanks for the help here. This is the 2nd issue I've seen relating to binding creation, and it might be an issue with the backwards compatibility of `x-azure-settings`. You should be able to bring everything up one level and remove that property entirely. But I'll investigate further when I get some time

PR #494, opened 2020-08-28
feat: Allow custom app insights instrumentation key

tbarlow12:
## What did you implement:

Enable custom app insights instrumentation key 

```yaml
provider:
  appInsights:
    instrumentationKey: {your_key}
```

## How did you implement it:

If an instrumentation key is added, arm template will use that value instead of the key from the generated template

**Out of scope**: The default template will still create an instance of app insights - should be future PR to remove that

## How can we verify it:

Deploy a function app with a custom app insights instrumentation key

## Todos:

       **Validate via `npm test`**
- [x] Write documentation


src/armTemplates/resources/functionApp.ts
```diff
@@ -221,7 +224,7 @@ export class FunctionAppResource implements ArmResourceTemplateGenerator {
       },
       {
         name: "APPINSIGHTS_INSTRUMENTATIONKEY",
-        value: "[reference(concat('microsoft.insights/components/', parameters('appInsightsName'))).InstrumentationKey]"
+        value: instrumentationKey || "[reference(concat('microsoft.insights/components/', parameters('appInsightsName'))).InstrumentationKey]"
```
wbreza, line 227: Unit test to validate this change?

